Issue #16, opened 2015-03-20
Fix check_pgactivity to run with Perl 5.10.0


frost242:
The line "use v5.8.0;" prevents check_pgactivity from working with Perl 5.10.0. It's obviously a bug as related in : http://www.nntp.perl.org/group/perl.perl5.porters/2008/01/msg132524.html

Adding use 5.006; before this line fix the issue.

PR #17, opened 2015-03-23
Fix check_pgactivity to run with Perl 5.10.0

frost242:
Hello,
This should fix issue #16 
Regards


ioguix:
As discussed by chat, I cancel this pull request for the next one.

Thank you!

PR #33, opened 2016-03-15
password support

michalmedvecky:
I needed to send password as well.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -236,6 +236,10 @@ Database server host or socket directory (default: "localhost").
 
 Database user name (default: "postgres").
 
+=item B<--password> PASSWORD
+
+Database password
+
 =item B<-p>, B<--port> PORT
 
 Database server port (default: "5432").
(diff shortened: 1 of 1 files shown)


ioguix:
I would recommend to use a service file, with proper access rights, instead of exposing a password to the world as command argument.
Did you try with a service file ?

michalmedvecky:
I did not, as there was no documentation for the plugin. I don't mind exposing the password as it runs on a dedicated machine only. But I understand your concern; you can add something like `($0 = "$0 @ARGV") =~ s/--password=\K\S+/x/;`

ioguix:
As far as I can see, service files are documented on the pgsql side AND check_pga side:
* http://www.postgresql.org/docs/9.5/interactive/libpq-pgservice.html
* https://github.com/OPMDG/check_pgactivity/blob/master/README.rst, see `-S, --dbservice SERVICE_NAME`

ioguix:
Hi,

I'm closing this PR as there was no activity for a long time and safe workaround exists using a service file or `.pgpass`.

PR #35, opened 2016-03-22
Add : table_unlogged service

ng-pe:
_Add : table_unlogged service_
Check if table are changed to unlogged. In 9.5+, you can switch between logged and unlogged. This service track change and a critical alert is raised if an unlogged table is detected.


ioguix:
I'll do some tiny changes in a next commit to: 
  * enforce a check on the 9.5 version and more
  * use a simple query instead of query_ver

Thank your for this patch!

PR #57, opened 2017-01-02
Fix a bug in the locks probe that occurs in PostgreSQL 8.2.

frost242:
Apply the same fix to all other versions for consistency, but the problem does not seem to occur on other versions.


ioguix:
Thank you!

PR #59, opened 2017-01-25
Add stat_snapshot_age to help monitor the current state of stats collector process.

frost242:
PostgreSQL 9.5 provides a function pg_stat_get_snapshot_timestamp() to provide the current statistics snapshot timestamp. This function can help to detect a stuck stats collector.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -237,7 +237,11 @@ my %services = (
     'pg_dump_backup' => {
         'sub'  => \&check_pg_dump_backup,
         'desc' => 'Check pg_dump backups age and retention policy'
-    }
+    },
+    'stat_snapshot_age' => {
+        'sub'  => \&check_stat_snapshot_age,
+	'desc' => 'Check stats collector\'s stats age'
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -6372,6 +6376,74 @@ sub check_wal_files {
...
+    my @rs;
+    my @perfdata;
+    my @msg;
+    my @hosts;
+    my %args     = %{ $_[0] };
+    my $stats_age;
+
+    my %queries  = (
+    $PG_VERSION_95 => q{ SELECT extract(epoch from (pg_stat_get_snapshot_timestamp() - now())) AS age }
+    );
```
ioguix: why using `%queries` and `query_ver` when you have only one query for 9.5+? ISTM a simple query and using the `query()` sub would be enough.

frost242:
Change done in commit 707d23d422d1d960e6bdb69d32f79b2842270b8b

ioguix:
OK, thank you!

PR #60, opened 2017-01-26
Add handling of PostgreSQL version 10

frost242:
Modify code to handle the new PostgreSQL versionning schema.


ioguix:
about commit a74901500436617e52cb3a3f6c3c2f2ceb4d9a7f, it would be more simple to use "server_version_num" from pg_settings and fallback to current code when no rows comes back (pg < 8.2).

PR #61, opened 2017-01-26
Add basic support for PostgreSQL 10

frost242:
Use server_version_num when possible, fallback to server_version evaluation if not possible.


ioguix:
Thank you Thomas!

PR #68, opened 2017-02-02
New sequences_exhausted probe

marco44:
First try at a sequences_exhausted probe. The purpose it to make sure there is no auto-incremented column that will fail in the close future in a database because it has either reached the maximum value of the sequence or the maximum value of the data type.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -244,6 +244,10 @@ my %services = (
         'sub'  => \&check_stat_snapshot_age,
 	'desc' => 'Check stats collector\'s stats age'
     },
+    'sequences_exhausted' => {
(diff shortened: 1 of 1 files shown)


frost242:
Since PostgreSQL 10, a new pg_sequence catalog has been introduced, see commit 1753b1b027035029c2a2a1649065762fafbf63f3 in PostgreSQL.

Your patch seems OK, expect this change. We're preparing next release to support PostgreSQL 10, so we should take care of such issue now.

Thanks

marco44:
It's become much simpler in 10.0, though 10.0 breaks compatibility with previous versions. Fixed.

frost242:
Thanks, seems fine now. I think this patch is ready to merge.

PR #80, opened 2017-02-22
Check wal receiver

frost242:
Add check_wal_receiver probe.

This new probe behaves like check_streaming_replication, but on the slave's side.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -6779,6 +6779,14 @@ SELECT
 
 =item B<wal_receiver> (9.6+)
 
+Check replication status on standby's end.
+
+Perfdata returns wal receiver's pid, current lag and timeline.
+
+This service returns OK if wal receiver's status is 'streaming',
+returns OK if status is different but wal receiver is active, and
+return unknown if not connected.
+
 =cut
 
 sub check_wal_receiver {
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -6802,16 +6802,10 @@ sub check_wal_receiver {
 
     is_compat $hosts[0], 'wal_receiver', $PG_VERSION_96 or exit 1;
 
-    my %queries  = (
-	   $PG_VERSION_96 => qq{ SELECT pid, slot_name, status, received_tli,
-                            pg_xlog_location_diff(latest_end_lsn, received_lsn) AS lag
-                       FROM pg_stat_wal_receiver },
-	   $PG_VERSION_100 => qq{ SELECT pid, slot_name, status, received_tli,
-                            pg_wal_location_diff(latest_end_lsn, received_lsn) AS lag
-                       FROM pg_stat_wal_receiver },
-    );
+    my $query  = qq{ SELECT pid, slot_name, status, received_tli
+                       FROM pg_stat_wal_receiver };
 
-    @rs = @{ query_ver ( $hosts[0], %queries ) };
+    @rs = @{ query ( $hosts[0], $query ) };
 
     return unknown ( $me, ['No receiver connected'], \@perfdata )
         unless scalar @rs;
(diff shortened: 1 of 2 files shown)


check_pgactivity
```diff
@@ -6773,6 +6777,63 @@ SELECT
...
 }
 
+=item B<wal_receiver> (9.6+)
+
+Check replication status on standby's end.
+
+Perfdata returns wal receiver's pid, current lag and timeline.
+
+This service returns OK if wal receiver's status is 'streaming',
+returns OK if status is different but wal receiver is active, and
```
ioguix: Returns WARNING I suppose?

check_pgactivity
```diff
@@ -6773,6 +6777,63 @@ SELECT
...
+
+    $rs[0][5] =~ m{^([0-9A-F]+)/([0-9A-F]+)$};
+    my $sender_location = ($wal_size * hex($1) ) + hex($2);
+
+    $rs[0][3] =~ m{^([0-9A-F]+)/([0-9A-F]+)$};
+    my $recv_location = ($wal_size * hex($1) ) + hex($2);
+
+    my $delta = $sender_location - $recv_location;
+
+    # status can be stopped, starting, streaming, waiting, restarting, stopping
```
ioguix: I wonder if this comment is in the good location in the code?

check_pgactivity
```diff
@@ -6773,6 +6777,63 @@ SELECT
...
+
+    my $pid = $rs[0][0];
+    my $status = $rs[0][2];
+    my $recv_tli = $rs[0][4];
+
+    $rs[0][5] =~ m{^([0-9A-F]+)/([0-9A-F]+)$};
+    my $sender_location = ($wal_size * hex($1) ) + hex($2);
+
+    $rs[0][3] =~ m{^([0-9A-F]+)/([0-9A-F]+)$};
+    my $recv_location = ($wal_size * hex($1) ) + hex($2);
```
ioguix: use the xlog_diff internal function instead of computing the decimal yourself?

check_pgactivity
```diff
@@ -6773,6 +6777,63 @@ SELECT
...
+
+    return unknown ( $me, ['No receiver connected'], \@perfdata )
+        unless scalar @rs;
+
+    my $pid = $rs[0][0];
+    my $status = $rs[0][2];
+    my $recv_tli = $rs[0][4];
+
+    $rs[0][5] =~ m{^([0-9A-F]+)/([0-9A-F]+)$};
+    my $sender_location = ($wal_size * hex($1) ) + hex($2);
```
ioguix: As far as I understand the doc, this is the last location the standby reported as received to the master as a feedback). "Lag" might be understood as a "lag" between the master and the standby, but it's just between what has been reported what has been flushed on disk in reality.
I wonder what is the purpose to report this "lag in feedback" as perfdata...

check_pgactivity
frost242: No, it has nothing to do here. I'll move it elsewhere or remove it simply.

check_pgactivity
frost242: Good point. Version 9.6 supports this function already.

check_pgactivity
frost242: Yep. I also ask myself about the purpose of this specific check, if it is really useful. We already have check_streaming_delta to detect that a slave is not connected anymore.

I'll find out how developers have come to this view, this will possibly help to decide.

check_pgactivity
frost242: Fixed in commit 2fb8c22

frost242:
Commit 2fb8c22 fixes the documentation issue, commit 9906d43 addresses other concerns.

frost242:
As discussed on instant messaging, this probe is not that useful. It reports the timeline id on a slave, but we can also obtain it from the function pg_control_checkpoint().

We close this pull request for now. If one find this feature useful, don't hesitate to commit on the PR.

PR #82, opened 2017-02-23
Show wait events in check_backends_status

frost242:
Commit 58a5303b9b78eb7025840f87ab003cf4da486f82 introduced support
for the new PostgreSQL wait event features from 9.6+. However it only
displayed heavyweight locks and didn't care about other wait event types.

This commit fixes this issue by introducing a new perfdata 'other wait event'
for check_backends_status.

A more complete patch may come later to count wait events more precisely.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -1762,7 +1762,8 @@ sub check_backends_status {
         'active'                        => [0, 0],
         'disabled'                      => [0, 0],
         'undefined'                     => [0, 0],
-        'insufficient privilege'        => [0, 0]
+        'insufficient privilege'        => [0, 0],
+        'other wait event'              => [0, 0]
     );
     my %queries      = (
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -1917,6 +1919,9 @@ sub check_backends_status {
     delete $status{'idle in transaction (aborted)'}
         if $hosts[0] < $PG_VERSION_90;
 
+    delete $status{'other wait event'}
+        if $hosts[0] < $PG_VERSION_96;
```
ioguix: Congrats ! You probably found a bug in the existing check_pga code :)

It should probably be `$hosts[0]->{'version_num'}`, not just `$hosts[0]`

frost242: Wow, I didn't notice. I already used the good one in another patch. Fix is on the way.

PR #89, opened 2017-03-14
Do not return pg_basebackup replication in streaming_delta

moumoul:
I think, pg_basebackup should not be returned by streaming_delta because it not a "normal" replication steam.
We just launch pg_basebackup in particular cases.


rjuju:
Hello,

I agree that it'd be better to exclude pg_basebackup replication from the results of streaming_delta.  However, the problem is that your patch assume that all pg_basebackup connection will always have 'pg_basebackup' as application_name, and that no other standby will ever use it.

I think it'd be better to handle a -`-exclude` parameter for this check, and let user configure with as they want.

moumoul:
Thanks for your feedback.
I'm checking how to add --exclude parameter and will send a new pull request :)

rjuju:
Ok, great!

I'm closing this PR for now, waiting for the new one. Thanks.

PR #90, opened 2017-03-14
Add --exclude option in check_streaming_delta

moumoul:
As discussed in PR #89 , here the PR which add --exclude option available for streaming_delta service.

We can exclude every pattern we want from the sql result.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -5491,8 +5491,7 @@ Either of the two following examples will check for the presence of two slaves:
   --slave 'slave1 192.168.1.11','slave2 192.168.1.12'
 
 This service supports a C<--exclude REGEX>  parameter to exclude every result
-matching the given regular expression. This allows you to filter either on an
-application_name or address IP or whatever you want.
+matching the given regular expression on application_name or address ip fields.
 
 You can use multiple C<--exclude REGEX>  parameters.
 
@@ -5558,18 +5557,16 @@ sub check_streaming_delta {
 
     @rs = @{ query_ver( $hosts[0], %queries ) };
 
-    my $index = 0;
+    my $excluded = 0;
     if (@exclude){
         foreach my $line (@rs){
-	    foreach my $exclude_re (@exclude){
-	        if (grep { $_ =~ $exclude_re } @{$line}){
-	            delete $rs[$index];
+	        if ((grep { @{$line}[0] =~ m/$_/ } @exclude) || (grep { @{$line}[1] =~ m/$_/ } @exclude)){
+		        $excluded++;
 	        }
-	    }
-	    $index++;
-	}
+	     }
     }
-	
+
+	push @perfdata => [ '# of excluded slaves', $excluded ];
     push @perfdata => [ '# of slaves', scalar @rs || 0 ];
 
     return unknown( $me, ['No slaves connected'], \@perfdata )
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -5514,6 +5514,7 @@ sub check_streaming_delta {
     my @hosts;
     my %slaves;
     my %args            = %{ $_[0] };
+    my @exclude      = @{ $args{'exclude'} };
     my $wal_size        = hex('ff000000');
     my $me              = 'POSTGRES_STREAMING_DELTA';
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -5551,6 +5558,18 @@ sub check_streaming_delta {
 
     @rs = @{ query_ver( $hosts[0], %queries ) };
 
+    my $index = 0;
+    if (@exclude){
+        foreach my $line (@rs){
+	    foreach my $exclude_re (@exclude){
+	        if (grep { $_ =~ $exclude_re } @{$line}){
```
rjuju: I'd prefer to explicitely test the regexp on the application_name and client_addr fields (and update the related comment at the beginning of this function).

check_pgactivity
```diff
@@ -5551,6 +5558,18 @@ sub check_streaming_delta {
 
     @rs = @{ query_ver( $hosts[0], %queries ) };
 
+    my $index = 0;
+    if (@exclude){
+        foreach my $line (@rs){
+	    foreach my $exclude_re (@exclude){
```
ioguix: I would rather move this loop directly in the beginning of the loop at line 5587 and replace the `delete` by te following code:

```
next if grep { $_ =~ $exclude_re } @{ $host }
```

At least, it would follow the same template code than other services using --exclude.

Thoughts?

check_pgactivity
```diff
@@ -5551,6 +5558,18 @@ sub check_streaming_delta {
 
     @rs = @{ query_ver( $hosts[0], %queries ) };
 
+    my $index = 0;
```
rjuju: I think it'd be better to handle the excluded connections in the "foreach my $host (@rs) {" loop.

I guess you did it this way to get the number of non excluded connections, but I think the "# of slaves" perfdata should return the real number of connections.

moumoul: If you keep "#of slaves" with the excluded connections then what's the point to this change ?
I think if we take the decision to exclude something it should be counted out of total of connected slaves.

check_pgactivity
moumoul: The total of slaves is calculated on line 5573

`push @perfdata => [ '# of slaves', scalar @rs || 0 ];
`

If I exclude my pattern on line 5587 The number '#of slaves' will differ from the number of returned lines.
My goal is to exclude the pattern from every results.

check_pgactivity
moumoul: Yes, it makes sens. I'll fix that asap.

check_pgactivity
rjuju: I thought the point of this patch was to avoid having false positive warning for outdated streaming connections due to pg_basebackup.

Having the "#of slaves" with the excluded connections allows to monitor if you exhaust the total number of possible connections for instance, and possibly know when/how long temporary use of streaming connections happened.

moumoul: I see. Maybe add a new value like "#of temp slave" with only matched value from parameters could be a solution ?

rjuju: That could work for me, maybe with a better name (like excluded slaves or something).

To be clear, you propose to have #of slaves + #of excluded slave = number of walsenders?

@ioguix What's your opinion on it?

moumoul:
Yes, this is what I propose.

Le jeu. 16 mars 2017 à 15:43, Julien Rouhaud <notifications@github.com> a
écrit :

moumoul:
Hi,

I've change the code to only count the number of excluded connexions.
# of slaves is the real number of connexions
# of excluded slaves is the number of excluded connexion. With that value we can do the difference between the real number and the number of excluded connexions for dynamic threshold in monitoring.

ioguix:
Sorry for the delay to handle this...

I'll do some cosmetic changes, but I'm fine with the PR now.

Thank you for your patch @moumoul!

ioguix:
Wait,

While doing my cosmetic changes, I just realized the last commit actually broke your code. You do not remove the excluded record anymore, just increment the `$excluded` var :/

I'll fix this by myself.

ioguix:
Fixed.

Please, see, comment and test c89af26.

Cheers,

PR #94, opened 2017-03-22
make it possible to run check_pgactivity without connection parameters

tobixen:
ref https://github.com/OPMDG/check_pgactivity/issues/92

Disclaimer: I haven't programmed perl for more than ten years, and I haven't tested this thoroughly yet.

Issue #92, opened 2017-03-22
Default host


tobixen:
I want  to be able to use `/usr/local/lib/monitoring/plugins/check_pgactivity -s $service` without any connection parameters - and it should initiate psql without connection parameters.

The default behaviour for psql is to connect through the unix socket file, found at the compile-time configured default socket directory (/tmp in the official postgres release, but /var/run/postgres under most distributions) and then connect as the logged-in-user (I'll add an nrpe user with superuser privileges, alternatively just do `sudo -u postgres`).

Two things:

* It's required to add either at least one connection parameter.  This is just an aesthetic issue, I just don't want to pass a "dummy" connection parameter.

* The default host is "localhost".  This causes a TCP connection, and the user is no longer authenticated by the OS, so one is asked for a password:

    $ sudo -u postgres psql -h localhost
    Password: 
    psql: fe_sendauth: no password supplied

Possibly this could be solved by running up an ident server, but I don't want to go that route.

I could add '--host /tmp' or '--host /var/run/postgres', but I don't like to do such hard-coding as it may be distro-specific where the socket file is located.

I could add some config service file on the host with a password ... but I don't want to complicate the setup.

Last, `-h ""` seems to work, but it does give some warnings so perhaps best to avoid it:

    $ sudo -u postgres /usr/local/lib/monitoring/plugins/check_pgactivity -h '' -s autovacuum
    Use of uninitialized value in sprintf at /usr/local/lib/monitoring/plugins/check_pgactivity line 879.
    POSTGRES_AUTOVACUUM OK: Number of autovacuum: 0 | ANALYZE=0 VACUUM=0 VACUUM_FREEZE=0 VACUUM_ANALYZE=0 oldest_autovacuum=NaNs max_workers=3

I'm considering writing up a pull request on this one.

maethor:
@tobixen you can use `-h /var/run/postgresql` without other user and password params.

tobixen:
> @tobixen you can use `-h /var/run/postgresql` without other user and password params.

Ref the research I did in march 2017, even though /var/run/postgresql is the correct answer in most distributions, the default directory in the official postgresql release is (or, was at that time) /tmp.

I don't want to hard code /var/run/postgresql into my manifests as it can be fragile.

ioguix:
Fixed in #213, thank you!

PR #96, opened 2017-03-23
make it possible to run check_pgactivity without connection parameters.

tobixen:
Redoing pull request #94, fix for issue #92

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -882,6 +882,10 @@ sub parse_hosts(\%) {
         }
     }
 
+    if (!@hosts) {
+	push(@hosts, {'name' => 'localhost/default'});
+    }
+
     dprint ('Hosts: '. Dumper(\@hosts));
 
     return \@hosts;
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -508,9 +508,10 @@ sub save($$$$) {
 
     if (defined $host->{'dbservice'}) {
         $hostkey = "$host->{'dbservice'}";
-    }
-    else {
+    } elsif (defined $host->{'host'}) {
         $hostkey = "$host->{'host'}$host->{'port'}";
+    } else {
+	$hostkey = "localhost/default"
     }
 
     die "File «${storage}» not recognized as a check_pgactivity status file.\n\n"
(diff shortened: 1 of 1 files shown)


Krysztophe:
I've tested it, at first sight it seems to work.

`./check_pgactivity --service database_size  -w 10% -c 20%` connects locally (/var/run/postgresql)

but it introduces an inconsistency, as `./check_pgactivity --service database_size -p 5432 -w 10% -c 20%` still wants to connect through localhost.

And the --help still says that it's defaulting to localhost.

tobixen:
Hm, when running "psql --port 5432" (or any other port number) without specifying a host, it apparently still attempts to connect to the socket - so to be consequent with that, we should simply ignore the port number if it's given without a host.  At the other hand, I think "localhost" is a sane default if a port number is given, hence IMO the only thing that is needed is the docfix (as in commit 866543f)

I got lots of conflicts when trying to do a "git rebase", perhaps I should try to redo this patch manually and submit a new pull request?

rjuju:
I'm not sure that I understand your point.  You can specify a specific port when connecting on a unix socket.  You also can't assume that check_pgactivity will use a psql binary compiled with default port, so for some people --port 5432 might not be the default behavior.


Otherwise, yes we need a clean branch to merge.  If that's easier for you, feel free to start a new PR instead.

tobixen:
@Krysztophe commented it's inconsistent to connect to localhost by default if a port number but no host is given, but to the unix domain socket if neither host nor port number is given.  I do disagree on that, but it's no big deal, I'll fix whatever it takes to get the pull request approved, my setup depends on this one.  (and yes, I'll fix a clean branch forked from the master branch)

The reason why I disagree ... I've never heard about usage of port numbers towards a unix domain socket.  If one has two different postgres instances running on one host, one would use two different unix sockets for them rather than two different port numbers on the same unix socket.  I seriously believe --port is a moot option if one is connecting to postgres over a socket.  For backward compatibility I believe it's better to default to "localhost" than the default unix socket if port number is given but not host name.

rjuju:
Yes, of course you'll endup with different unix sockets if you have multiple instances.  What I meant it that libpq will connect to a specific socket name in the given path depending on the port you requested, so you have to keep the port information even if you have a unix socket path.

PR #100, opened 2017-04-07
Some more PostgreSQL 10 fixes

frost242:
This bunch of commits take account of the new pg_monitor role in PostgreSQL 10.

Using this feature, we'll be able to use an unprivileged monitoring user. The following services were changed : 
  * wal_files
  * archiver

The btree_bloat and table_bloat services still works but one needs to grant the SELECT privilege on table pg_statistic to pg_monitor, inside each monitored databases, otherwise the check will fail. This is of course documented.

Also, one generic fix concerns the backends_status service. The query column was retrieved but was not used at all. Now this service simply retrieves actually used columns.

### DIFF
--- a/README
+++ b/README
@@ -343,6 +343,10 @@ check_pgactivity
         perfdata. This list contains the fully qualified bloated index name,
         the estimated bloat size, the index size and the bloat percentage.
 
+        This service will work with PostgreSQL 10+ without superuser privileges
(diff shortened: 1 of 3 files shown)


ioguix:
Just quick note: add some doc for each service about the superuser requirement.

frost242:
Didn't think about it while talking about it irl, it's already addressed in commit 1fe3945

PR #108, opened 2017-06-11
Integration with Travis CI

daamien:
This patch contains 3 main changes:

* A ``test.sh`` file that runs basic check_pgactivity lines and stops if there's an error. The file is not complete yet, this is just a PoC. The script takes the major version as parameter

* A ``dockertest.sh`` that launches the ``test.sh`` inside any official docker postgres image (see https://hub.docker.com/_/postgres/ ).  The script takes the major version as parameter

* A ``Travis CI config file`` that launches ``dockertest.sh`` on every major version from 9.2 to 9.6. (the postgres 10 image is on its way)

As a result, any developper can create a Travis CI account, activate integration for the check_pgactivity repository and then every new commit will be checked like this :

https://travis-ci.org/daamien/check_pgactivity/builds/241807277

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+sudo: required
+
+services:
+  - docker
+
+script:
+- ./dockertest.sh 9.2
+- ./dockertest.sh 9.3
+- ./dockertest.sh 9.4
+- ./dockertest.sh 9.5
+- ./dockertest.sh 9.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+XXXX-XX-XX WIP:
+  - add ``test.sh`` and ``dockertest.sh``
+	- add TravisCI integration
+  - add a ``CONTRIBUTING`` file
+
+
 2017-04-28 v2.2:
   - add support for PostgreSQL 9.6
   - add early-support for PostgreSQL 10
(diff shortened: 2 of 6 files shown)


ioguix:
Hello,

Thank you for this (old) PR and sorry for the lack of feedback. I needed some time and more experience in testing env before being able to review your PR.

In the process of next release, I started working on this subjecta and had a look at your PR few weeks ago. There's some pros and some cons. In short:

  * we might want to use Travis in short/mid term
  * Travis allows to test on various OS, including ubuntu
  * we can test check_pga against many different packaged version of PostgreSQL on Ubuntu
  * no need or very little need of scripting with Travis/Ubuntu to setup the test environment
  * Docker has fewer PostgreSQL available and requires (at first glance) more scripting to setup the test environment
  * your testing script doesn't seem to be the best fit as we have a **LOT** to test in check_pga, not just the success return code. We need to tests error code as well, plus stdout, stderr, perdata and so on.

I started a branch on my repo to create tests based on the standard TAP tests protocol using the Perl modules from PostgreSQL project. See folder `t/` in https://github.com/ioguix/check_pgactivity/tree/tests . Perl modules from PostgreSQL are really powerful and I'm quite glad to be able to deeply tests check_pga with so few lines of code thanks to them. As far as I understand it, this should perfectly fit in Travis as well.

I'll keep feeding my branch and will create a WIP PR soon to prepare for a merge in next release 2.6. You'll be welcome to join the effort in regard with Travis there if you agree and have time.

Thank you!

PR #109, opened 2017-07-07
Fix fetching method

loukash:
Specific tool didn't use for fetch latest version online.
_Curl_ didn't follow to redirect if option "--location" not set (http redirect to https).


ioguix:
Hi,

The patch makes sense and looks good.

Thanks for the fix and cleanup!

PR #123, opened 2017-10-06
Update comments for check_wal_files for PG >=9.5

Krysztophe:
Added missing comment on the source of data for check_wal_files and PG>=9.5


ioguix:
Good catch, thank you for your patch!

PR #130, opened 2017-11-06
[Doc] Some services always exclude templates and the posgres database

Krysztophe:
It was not clear from the doc that the postgres db and templates are never checked by some services : added a sentence for btree/table_bloat, invalid_indexes, last_analyze/vacuum, table_unlogged, sequences_exhausted


frost242:
Thanks !!

PR #132, opened 2017-11-07
Bug minor version in v10

Krysztophe:
Symptom : ``POSTGRES_MINOR_VERSION UNKNOWN: Unknown major PostgreSQL version 10.0``
It fails both with and without the --critical switch. In both cases the x.y.z  structure is still hard coded to parse the RSS file or the threshold.

I've found no other way than making different tests if version is after or before 10.

Works on my test cases. I'll check when 10.1 is released this week.

(Warning: perl beginner code!)

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -4564,13 +4564,14 @@ seperated by anything but a '.'.
 
 For instance, the following parameters are all equivalent:
 
-  --critical "9.3.2 9.2.6 9.1.11 9.0.15 8.4.19"
-  --critical "9.3.2, 9.2.6, 9.1.11, 9.0.15, 8.4.19"
-  --critical 9.3.2,9.2.6,9.1.11,9.0.15,8.4.19
-  --critical 9.3.2/9.2.6/9.1.11/9.0.15/8.4.19
+  --critical "10.1 9.6.6 9.5.10 9.4.15 9.3.20 9.2.24 9.1.24 9.0.23 8.4.22"
+  --critical "10.1, 9.6.6, 9.5.10, 9.4.15, 9.3.20, 9.2.24, 9.1.24, 9.0.23, 8.4.22"
+  --critical "10.1,9.6.6,9.5.10,9.4.15,9.3.20,9.2.24,9.1.24,9.0.23,8.4.22"
+  --critical "10.1/9.6.6/9.5.10/9.4.15/9.3.20/9.2.24/9.1.24/9.0.23/8.4.22"
 
-Any value other than 3 numbers separated by dots will be ignored.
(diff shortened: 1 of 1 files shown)


Krysztophe:
Just tested: 
After the 10.1 was released, the service on 10.0 instance is critical. After the upgrade to 10.1, it finds 10.1 correctly and returns OK again.
Tested with the RSS feed and the --critical parameters.
Tested on 9.6.5 instances and others which behave like before.
Does not recognize a 11devel version, but this is not a problem.

Krysztophe:
Rebased & squashed into one commit

check_pgactivity
```diff
@@ -4564,13 +4564,14 @@ seperated by anything but a '.'.
...
-  --critical 9.3.2/9.2.6/9.1.11/9.0.15/8.4.19
-
-Any value other than 3 numbers separated by dots will be ignored.
-if the running PostgreSQL major version is not found, the service raises an
+  --critical "10.1 9.6.5 9.5.9 9.4.14 9.3.19 9.2.23 9.1.24 9.0.23 8.4.22"
+  --critical "10.1, 9.6.5, 9.5.9, 9.4.14, 9.3.19, 9.2.23, 9.1.24, 9.0.23, 8.4.22"
+  --critical "10.1,9.6.5,9.5.9,9.4.14,9.3.19,9.2.23,9.1.24,9.0.23,8.4.22"
+  --critical "10.1/9.6.5/9.5.9/9.4.14/9.3.19/9.2.23/9.1.24/9.0.23/8.4.22"
+
+Any value other than 3 numbers separated by dots (before version 10.x)
```
rjuju: whitespace here

check_pgactivity
```diff
@@ -4646,8 +4647,13 @@ sub check_minor_version {
...
             unless $rss;
 
+        # versions until 9.6.x
         $latest_versions{"$1.$2"} = [$1 * 10000 + $2 * 100 + $3, "$1.$2.$3"]
-            while $rss =~ m/<title>(\d+)\.(\d+)\.(\d+)/g;
+            while $rss =~ m/<title>(\d+)\.(\d+)\.(\d+)/g && $1<10;
+        # versions from 10
+        $latest_versions{"$1"} = [$1 * 10000 + $2, "$1.$2"]
+            while $rss =~ m/<title>(\d+)\.(\d+)/g && $1>=10;
+
```
rjuju: could you add parens around the while, it'll make the code more readable

rjuju: I meant this one and the one before of course

Krysztophe:
Rebase KO, je refais une branche et une PR.

PR #134, opened 2017-11-10
Minor version compatible with v10

Krysztophe:
Symptom : ``POSTGRES_MINOR_VERSION UNKNOWN: Unknown major PostgreSQL version 10.0``
It fails both with and without the --critical switch. In both cases the x.y.z  structure is still hard coded to parse the RSS file or the threshold.

I've found no other way than making different tests if version is after or before 10.

Test:
After the 10.1 was released, the service on my 10.0 instance is critical. After the upgrade to 10.1, check_pga finds 10.1 correctly and returns OK again.
Tested with the RSS feed and the --critical parameters.
Tested on 9.6.5 instances and others which behave like before.
Does not recognize a 11devel version, and should not.

Includes small modifications that rjuju asked for.

(2nd PR as I've messed up the last rebase)

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -4564,13 +4564,14 @@ seperated by anything but a '.'.
 
 For instance, the following parameters are all equivalent:
 
-  --critical "9.3.2 9.2.6 9.1.11 9.0.15 8.4.19"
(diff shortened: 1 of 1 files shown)


frost242:
Thanks !

PR #138, opened 2017-11-13
archive_folder : doc update to clarify where to execute

Krysztophe:
archive_folder: added a line to make it clear that it must be executed on a server with access to the archive_folder, and this is not always the instance server.
Implicit mais ça va mieux en le disant.


frost242:
Thanks

PR #140, opened 2017-11-16
Fix exclude option in backends_status service

loukash:
Exclude options are ignored. It's usefull option if you want to exclude "ANALYZE" or "VACUUM" queries.
In current version "$r->[2]" is current_setting('max_connections'). Isn't query string.


frost242:
Thanks for this fix !

frost242:
Btw, it will be included in next release. The current one is still pending some announcement.

I hope to make a new release in the next two months schedule.

Issue #106, opened 2017-06-08
Support pgbouncer


frost242:
Add support for some pgbouncer services.

frost242:
Support for pgbouncer will take place in another monitoring script. Question is open about hosting check_pgbouncer in the same source tree or make a new projet out of it.

anayrat:
Useful posts as reminder : 

- https://blog.okmeter.io/use-red-and-real-world-pgbouncer-monitoring-61b34ebeebb8
- https://blog.okmeter.io/pgbouncer-monitoring-improvements-in-recent-versions-5905f5d1a1ed

PR #144, opened 2017-11-30
[Do not merge] - Pgbouncer services

anayrat:
This PR should address #106 and other things (typo, sort services).

TODO : 

  * [x] : Doc
  * [x] : Pgbouncer stats service (just to collect stats)
  * [x] : handle when max_db_connection is not present in database section (pgbouncer use max_db_connection from default config). We must join with SHOW CONFIG
  * [x] : sometime pgbouncer clean database from show database. If we join with show pools, result is not consistent. pool are present but databases are missing.

Feature request : 

  * [x] : output without perfdata : some services provide serveral perfdata per pool or database. It generates plenty perfdatas when the cluster has hundred databases. https://github.com/OPMDG/check_pgactivity/issues/156

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ You must provide some mandatory data in the @perfdata array :
 
 * perfdata name
 * the data itself, in numerical form
-* the unit used:'B' for bytes, 's' for secondes or undef for raw numbers
+* the unit used:'B' for bytes, 's' for seconds or undef for raw numbers
 
 The following data are optional :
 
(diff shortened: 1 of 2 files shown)


anayrat:
Reviews are welcome!

anayrat:
Erf, I notice pgbouncer 1.8 change columns order in statistics view. This PR can only work with pgbouncer 1.7.* :(

I wonder if we should create a separate project with dedicated functions for each view according to pgbouncer's version.
It avoid duplicated code, allow to handle different pgbouncer's versions and avoid to bloat check_pgactivity.

frost242:
Should we close this pull request ?

frost242:
Well, I shouldn't have asked the question, issue #106 indicates that pgbouncer support will be held in another script or project.

PR #175, opened 2018-03-16
Add required privileges to uptime service comments

Krysztophe:
...as this service was not merged when I was adding all the comments about privileges


frost242:
Thanks !

PR #178, opened 2018-04-02
WIP:Relation size

Krysztophe:
Add a relation_size service to follow the size of a few tables, including aggregated tables.

By default, it lists all tables, indexes, materialized views.
--include 'tablename' lists one (ore more) relations and all its(their) indexes and (sub)partitions with their pg_relation_size
--agg aggregates a table and its partitions (both partitioning modes) including indexes and toast.
--details shows vm,fsm,main,toast,idx of tables instead of just pg_relation_size (and aggregates the indexes)
--agg --details shows vm,fsm,main,toast,idx summed up for all partitions.

Warning and Critical (optional) are triggered by size (compared to each returned line).

This service is fully functional but may need more work: the main query is a bit tormented and may be simplified or tuned, the perl part is a bit naive. The main query uses dynamic filtering to reduce the load on the server. Complete query time is usually below the second on my laptop, even with thousands of tables, but may need a few seconds in degenerated cases (10'000 partitions); and much more time is spent in check_pgactivity itself if --agg was omitted (but listing thousands of tables pointless).

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -117,7 +117,6 @@ my %services = (
         'sub'  => \&check_table_unlogged,
         'desc' => 'Check unlogged tables'
     },
-
     'wal_files' => {
         'sub'  => \&check_wal_files,
         'desc' => 'Total number of WAL files.',
@@ -219,6 +218,10 @@ my %services = (
     'pga_version' => {
         'sub'  => \&check_pga_version,
         'desc' => 'Check the version of this check_pgactivity script.'
+    },
+    'relation_size' => {
+        'sub'  => \&check_relation_size,
+        'desc' => 'Check relation sizes.'
     },
     'is_replay_paused' => {
         'sub'  => \&check_is_replay_paused,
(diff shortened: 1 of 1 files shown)


Krysztophe:
Example with OPM:
table_classique_par_fork
table_classique_par_objets
table_partitionnee_par_forks
table_partitionnee_par_objets
table_partitionnee_total

Krysztophe:
Rebase done

Krysztophe:
Cleaning

To recreate from scratch later

ioguix:
There is so much different usages depending on the database, the number of relations, schema, etc. Should we use explicit `--include`/`--exclude` ? Should we return a top 10? Should we support regex? Should we report all relations? Shoult it raise alarm? On what?

This would be either a very large, complex, check, already highly customized command, close to what we could do with the custom query check, or a very limited and useless check.

Currently, I suppose this could be done using the custom query service.

PR #182, opened 2018-04-12
 Fix uptime failure to detect a crash-and-restart after a child process failure (10+)

Krysztophe:
After a crash and restart, the postmaster uptime stays the same although practically everything has restarted.
On PG10+, uptime is the age of the latest child process (usually checkpointer, writer...).
 pg_monitor at least is needed. Client backends are excluded because some were met that were older although dead.

The same fix on PG 9.6 and before will need to find another way.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7373,7 +7373,7 @@ sub check_uptime {
 
     return critical ( $me, \@msg_crit, \@perfdata) if ( @msg_crit ) ;
     return warning  ( $me, \@msg_warn, \@perfdata) if ( @msg_warn ) ;
-    return ok( $me, [ $msg_uptime $msg_reload_conf $msg_shmem_init_time ], \@perfdata );
+    return ok( $me, [ $msg_uptime, $msg_reload_conf, $msg_shmem_init_time ], \@perfdata );
 }
 
 =item B<wal_files> (8.1+)
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7255,18 +7255,27 @@ DB_LOOP: foreach my $stat (@rs) {
 
 =item B<uptime> (8.1+)
 
-Returns time since postmaster start (from 8.1) and configuration reload (from 8.4).
+Returns time since postmaster start ("uptime", from 8.1),
+since configuration reload (from 8.4),
+and since shared memory initialization (from 10).
+
+Please note that the uptime is unaffected when the postmaster resets
+all its children (for example after a kill -9 on a process or a failure).
+
+From 10+, the 'time since shared memory init' aims at detecting this situation:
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -7255,18 +7255,27 @@ DB_LOOP: foreach my $stat (@rs) {
...
-Warning is raised when the configuration uptime is less than the warning thereshold.
+raised when the postmaster uptime or the time since shared memory initialization
+is less than the critical threshold.
+Warning is raised when the time since configuration reload is less than the warning threshold.
 If only a warning or critical threshold is given, it will be used for both cases.
 Obviously these alerts will disappear from themselves once enough time has passed.
 
-Perfdata contain the duration since postmaster start and duration since
-configuration reload.
+Perfdata contain the three values, or NULL if not available for this version.
```
ioguix: Perfdata should not include data not supported on the hosts. Even if set to NULL.
```suggestion
Perfdata contains each values depending of the postgresql version and their availability.
```

check_pgactivity
```diff
@@ -7311,23 +7332,37 @@ sub check_uptime {
     $w_limit = get_time $args{'warning'}  if (defined $args{'warning'});
 
     @rs = @{ query_ver( $hosts[0], %queries ) };
-    # uptime since postmaster start
     $uptime = int( $rs[0][0] );
-    $msg_uptime = "instance started for ".to_interval($uptime)."s (since $rs[0][2])" ;
+    $msg_uptime = "postmaster started for ".to_interval($uptime)."s (since $rs[0][2])" ;
```
ioguix: `to_interval` already set the unit. 
```suggestion
    $msg_uptime = "postmaster started for ".to_interval($uptime)." (since $rs[0][2])" ;
```

check_pgactivity
```diff
@@ -7311,23 +7332,37 @@ sub check_uptime {
...
     } else {
         $msg_reload_conf = "";
     };
 
-   if ( defined $c_limit and $uptime < $c_limit ) {
+    # time since last share memory reinit
+    if ( check_compat $hosts[0], $PG_VERSION_100 ) {
+	  $shmem_init_time = int ( $rs[0][4] );
+	  $msg_shmem_init_time = "shared memory initialized since ".to_interval($shmem_init_time)." (since $rs[0][5])";
+	  push @perfdata => [ 'shmem reinit time', $shmem_init_time , 's',
```
ioguix: should be "init time", not "reinit"
```suggestion
	  push @perfdata => [ 'shmem init time', $shmem_init_time , 's',
```

check_pgactivity
Krysztophe: Corrected in my  branch

PR #186, opened 2018-04-24
Add a new service called extensions_versions

Krysztophe:
Check all extensions installed in all databases
and raise an alarm if the current version is not the default version available
on the instance.

Typically, it is used to detect forgotten extension upgrades after package
upgrades or a pg_upgrade.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -3915,9 +3915,17 @@ upgrades or a pg_upgrade.
 Perfdata returns the number of outdated extensions in each database.
 
 This service supports both C<--dbexclude> and C<--dbinclude> parameters.
+Schemas are ignored, as an extension cannot be installed more than once
+in a database.
 
 This service supports multiple C<--exclude> argument to exclude one or
-more extensions from the check.
+more extensions from the check. To ignore an extension only in a particular database,
+use  'dbname/extension_name' syntax.
+
+Examples:
+
+    --dbexclude 'devdb' --exclude 'testdb/postgis' --exclude 'testdb/postgis_topology'
+    --dbinclude 'proddb' --dbinclude 'testdb'  --exclude 'powa'
 
 Required privileges: unprivileged role able to log in all databases
 
@@ -3962,7 +3970,7 @@ sub check_extensions_versions {
 
         REC_LOOP: foreach my $ext (sort @rs) {
             foreach my $exclude_re ( @{ $args{'exclude'} } ) {
-                next REC_LOOP if $ext->[0] =~ /$exclude_re/;
+                next REC_LOOP if $ext->[0] =~ /$exclude_re/ or "$db/$ext->[0]" =~ /$exclude_re/ ;
             }
 
             $outdated++;
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -3906,7 +3906,7 @@ sub check_database_size {
 =item B<extensions_versions> (9.1+)
 
 Check all extensions installed in all databases (including templates)
-and raise a Critical alamr if the current version is not the default
+and raise a critical alert if the current version is not the default
 version available on the instance (according to pg_available_extensions).
 
 Typically, it is used to detect forgotten extension upgrades after package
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -3487,6 +3490,95 @@ sub check_database_size {
...
+Check all extensions installed in all databases
+and raise an alarm if the current version is not the default version available
+on the instance.
+
+Typically, it is used to detect forgotten extension upgrades after package
+upgrades or a pg_upgrade.
+
+Perfdata returns the number of outdated extensions in each database.
+
+The 'postgres' database and templates are always excluded.
```
rjuju: I don't think we should exclude any database by default.  It's especially important for template databases if people relies on the template databases to be up to date to create new production databases.

check_pgactivity
```diff
@@ -3487,6 +3490,95 @@ sub check_database_size {
...
+
+Critical and Warning thresholds accept numbers of outdated extensions.
```
rjuju: I'm not sure that a raw number for outdated extensions has any sense.  I think we should always returns critical if there's at least one outdated extension, and provide a way to exclude an extension, and a (database, extension) from the check.

check_pgactivity
Krysztophe: I agree.
I've just used get_all_dbname, like many other services. Perhaps it should get a parameter to avoid such filtering.

check_pgactivity
Krysztophe: I can imagine a case where you know that an extension will be outdated for a long time, will acknowledge a warning, and do not want to exclude the extension definitely.

So I can modify it this way:
- default 0 for critical and warning
- add --exclude parameter

rjuju: Acknowledging a warning would assume that you setup the critical threshold just above the number of expected outdated extension to get another notification if another one is outdated, and hope that an extension isn't updated while a new one is available.  Since it requires specific nagios configuration, it's less bug prone the define the explicit list of cases to ignore.

check_pgactivity
Krysztophe: Solved by PR #224

Krysztophe:
I've rebased and updated the branch:

- removed warning;  any obsolete extension leads to Critical
- all dbs and templates are tested
- I didn't add an  --exclude yet, I'm wondering if this overkill (there is already --dbexclude)

Sample:
```
$ ./check_pgactivity --service extensions_versions  -p9603 
POSTGRES_CHECK_EXT_VERSIONS CRITICAL | powa=3
powa.powa: 3.1.0 (should be: 3.2.0)  powa.pg_stat_kcache: 2.0.3 (should be: 2.1.1)  powa.pg_qualstats: 1.0.2 (should be: 1.0.7) 

./check_pgactivity --service extensions_versions  -p9603 --dbexclude powa
POSTGRES_CHECK_EXT_VERSIONS OK

./check_pgactivity --service extensions_versions -h paddington -p 11001
POSTGRES_CHECK_EXT_VERSIONS CRITICAL | christ=3 testpostgis=3                                                                
christ.postgis_topology: 2.5.1 (should be: 2.4.6)  christ.postgis: 2.5.1 (should be: 2.4.6)  christ.postgis_tiger_geocoder: 2.5.1 (should be: 2.4.6)  testpostgis.postgis_tiger_geocoder: 2.5.1 (should be: 2.4.6)  testpostgis.postgis: 2.5.1 (should be: 2.4.6)  testpostgis.postgis_topology: 2.5.1 (should be: 2.4.6) 
```

ioguix:
I just added a commit to your branch to add the --exclude argument, enforce schema prefix, always report perfdata, a small doc fix and fix indentation.

Could you please review?

Krysztophe:
@ioguix : Tests ok for `--exclude`... except that the schema is totally ignored.

In fact I forgot it totally in the main query.

I'm wondering if this would be really useful as you can only install an extension once in a db.

check_pgactivity
```diff
@@ -3572,6 +3575,85 @@ sub check_database_size {
...
+        next ALLDB_LOOP if grep { $db =~ /$_/ } @dbexclude;
+        next ALLDB_LOOP if @dbinclude and not grep { $db =~ /$_/ } @dbinclude;
+
+        my $outdated = 0;
+
+        # For each record: extension, default, installed
+        @rs = @{ query ( $hosts[0], $query, $db ) };
+
+        REC_LOOP: foreach my $ext (sort @rs) {
+            foreach my $exclude_re ( @{ $args{'exclude'} } ) {
```
rjuju: I still think that this is too simplistic.  What if for some reason you can't upgrade an extension in only one (or a subset) of your databases but want to check that it's up-do-date in all others?

ioguix: what do you propose as user interaction for such feature?

rjuju: Same as what I suggested in May, ie. a way to specify extname or (extname and dbname).  Maybe something like
 * `--exclude 'extname'`
 * `--exclude 'dbname/extname'`

ioguix: well, why not. I'll do it during the week.

Krysztophe: This PR is old enough, so I've added what seems to be the last missing part about `--exclude 'dbname/extname'`, and a few comments: 4f255cc25310e27915dceb896aa9a00997612862

Example:
```bash
$ check_pgactivity --service extensions_versions --dbservice=nagios_defo12 

POSTGRES_CHECK_EXT_VERSIONS CRITICAL | bis=2;;1;0 gin=2;;1;0 partitions_100_par_fdw=0;;1;0 postgres=2;;1;0 scratch=2;;1;0 template1=0;;1;0 textes10=0;;1;0 tpc=1;;1;0
bis.unit: 5 (should be: 7) bis.nulle: 0.0.2 (should be: 0.0.3) gin.nulle: 0.0.2 (should be: 0.0.3) gin.unit: 3 (should be: 7) postgres.unit: 3 (should be: 7) postgres.nulle: 0.0.2 (should be: 0.0.3) scratch.unit: 4 (should be: 7) scratch.nulle: 0.0.2 (should be: 0.0.3) tpc.nulle: 0.0.1 (should be: 0.0.3)

$ check_pgactivity --service extensions_versions --dbservice=nagios_defo12  --exclude 'scratch/nulle' --exclude 'bis/nulle' --exclude 'gin/nulle' --exclude 'unit'  --dbexclude postgres

POSTGRES_CHECK_EXT_VERSIONS CRITICAL | bis=0;;1;0 gin=0;;1;0 partitions_100_par_fdw=0;;1;0 scratch=0;;1;0 template1=0;;1;0 textes10=0;;1;0 tpc=1;;1;0
tpc.nulle: 0.0.1 (should be: 0.0.3)

$ check_pgactivity --service extensions_versions --dbservice=nagios_defo12  --exclude 'scratch/nulle' --exclude 'bis/nulle' --exclude 'gin/nulle' --dbexclude postgres --exclude 'tpc/nulle' --exclude 'unit'

POSTGRES_CHECK_EXT_VERSIONS OK | bis=0;;1;0 gin=0;;1;0 partitions_100_par_fdw=0;;1;0 scratch=0;;1;0 template1=0;;1;0 textes10=0;;1;0 tpc=0;;1;0
```

Krysztophe: Up! @rjuju are you okay?

ioguix: I did a code review and fix a small incompatibility.
The patch is ok to squash & merge imo.
@rjuju: one last review round/red flag before I merge?

rjuju: thanks a lot, it all looks good to me!

rjuju (approved):

Issue #194, opened 2018-07-24
wal_files:  in v10, wal_keep_segments is forgotten


Krysztophe:
According to the doc and the 9.5 code, the limit of number of wals is  `max_wal_size` (as segments)`+wal_keep_segments `.
The v10 version does not follow this logic, it just computes the number of wals from max_wal_size and makes no addition.

This seems to go back to c4950dd2

> For 9.5 and above, the limit is:
>   100% =  max_wal_size      (as a number of WAL)
>         + wal_keep_segments (if set)

Code for v9.5 and v10:
```
     $PG_VERSION_100 => q{
        WITH wal_settings AS (
          SELECT sum(setting::int) filter (WHERE name='max_wal_size') as max_wal_size,
                 sum(setting::int) filter (WHERE name='wal_segment_size') as wal_segment_size,
                 sum(setting::int) filter (WHERE name='wal_block_size') as wal_block_size,
                 sum(setting::int) filter (WHERE name='wal_keep_segments') as wal_keep_segments
          FROM pg_settings
          WHERE name IN ('max_wal_size','wal_segment_size','wal_block_size','wal_keep_segments')
        )
        SELECT s.name,
          max_wal_size / (wal_block_size * wal_segment_size / 1024^2) max_nb_wal,
          CASE WHEN pg_is_in_recovery()
            THEN NULL
            ELSE pg_current_wal_lsn()
          END,
          wal_keep_segments,
          (pg_control_checkpoint()).timeline_id AS tli
        FROM pg_ls_waldir() AS s
        CROSS JOIN wal_settings
        WHERE name ~ '^[0-9A-F]{24}$'
        ORDER BY
          s.modification DESC,
          name DESC},
     $PG_VERSION_95 => q{
        WITH wal_settings AS (
          SELECT setting::int + current_setting('wal_keep_segments')::int as max_nb_wal
          FROM pg_settings
          WHERE name = 'max_wal_size'
        )
        SELECT s.f,
          max_nb_wal,
          CASE WHEN pg_is_in_recovery()
            THEN NULL
            ELSE pg_current_xlog_location()
          END,
          current_setting('wal_keep_segments')::integer,
          substring(s.f from 1 for 8) AS tli
        FROM pg_ls_dir('pg_xlog') AS s(f)
        CROSS JOIN wal_settings
        WHERE f ~ '^[0-9A-F]{24}$'
        ORDER BY
[...]

PR #196, opened 2018-07-25
(v11 compatibility) wal_files and replications_slots

Krysztophe:
Modification of wal_files and replication_slots, as wal_segment_size is now stored as bytes, and not as blocks anymore.

Added a few comments to help the next one that will try to re-engineer this part.

Tested on a setup with 1 Go WAL files, it seems fine.

Wal_files assumes that #195 (bug #194) will be applied.


Krysztophe:
@ioguix : as discussed this morning, I've added a few comments

ioguix:
Thanks!

Issue #192, opened 2018-06-18
Add PostgreSQL 11 support


frost242:
This issue sums up the different changes introduced in PG11 that will need some work on our side.

- [x] First, issue #152 must be fixed.
- [ ] A new role must be documented : pg_read_server_files. This one fills the gap between pg10 and previous versions.
- ~~Allow access to file system functions to be controlled by GRANT/REVOKE permissions, rather than super-user checks (Stephen Frost)~~
- [x] Add pg_stat_activity.backend_type now shows the type of background worker (Peter Eisentraut)
- [x] Update XLOG_PAGE_MAGIC
- [x] handle wal_segment_size

I hope I have everything. This one is not as problematic as PG10.

anayrat:
FYI wal_segment_size unit has changed:
In PG 10:
```
name            | wal_segment_size
setting         | 2048
unit            | 8kB
```
In PG 11:
```
name            | wal_segment_size
setting         | 16777216
unit            | B
```
We will have to change several queries :(

rjuju:
@anayrat it changed in https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=fc49e24fa69

frost242:
Thanks, added to the list !

Krysztophe:
I've just pushed #196 about wal_segment_size. 

This corrects wal_files and replication_slots. At least archive_folder must be updated too.

Krysztophe:
I've pushed #197 for the minimal rights in v11. It seems that pg_reads_server_files is not very useful. The problems I've found for unprivileged users  are all linked to EXECUTE rights on pg_ls_dir and other functions.

frost242:
Hello,
Regarding the EXECUTE rights, does it also work with in version 10 ? I'm not able to test pg_read_server_files right now, I'll try to take a look at it soon. I hope to merge your PR in the coming days.

Krysztophe:
@frost242 : nope, on v10, you must be superuser to use pg_ls_dir(). I've tried and the error message is explicit.

frost242:
@Krysztophe Ah, I remember now. The pg_ls_dir() C code explicitely checks for a superuser role, in v10 and before !

ioguix:
#152 has been postoned to 2.5

frost242:
[...]

PR #198, opened 2018-08-06
Release 2.4 and other fix

anayrat:
Hi,

This PR contains:
- An example for replication_slot service
- A change for commit_ratio threshold from integer type to float. It may be useful when rollback_ratio is low (less than 1%), you can specify `rollback_ratio=0.1` for 0.1% rollback ratio.
- CHANGELOG.md updated for 2.4 release

I you prefer I can split this PR in separate PR for CHANGELOG.md vs other commits.

Regards,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,30 @@
 Changelog
 =========
 
+2018-18-XX v2.4:
+
+  - fix handling or file reading errors in archive_folder service
+  - fix wal magic number for version 10
+  - fix service stat_snapshot_age to output the correct age
+  - fix archiver and replication_slots services to work properly on a standby node
+  - fix is_replay_paused for PostgreSQL 10
+  - fix max_nb_wal calculation in wal_files service
+  - fix uninitialized bug in hit_ratio when database do not yet have statistics
+  - fix check_backend_status in order to ignore unknown status
+  - fix service sequences_exhausted to take account of sequence's minvalue
+  - fix sequences_exhausted to take account of sequences only in the current db
+  - fix exclude option in backends_status service
+  - ignore startup and backup replication states in service streaming_delta
+  - avoid warning for -dev versions in pga_version service
+  - add a new uptime service
+  - replication_slots service handle wal files and pg_replslots files separately
+  - add ability to filter by application_name in longest_query and oldest_xact service
+  - add minimal delta size to pgdump_backup service to avoid alert when backup grows small in size
+  - add documentation example for pgback in pgdump_service
+  - add documentation for archive_folder
+  - add privileges information for all services
+  - take account of the new BRIN summarize state of autovacuum
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,30 @@
 Changelog
 =========
 
+2018-18-XX v2.4:
```
rjuju: s/18/08/ :)

CHANGELOG.md
```diff
@@ -1,6 +1,30 @@
...
+  - fix uninitialized bug in hit_ratio when database do not yet have statistics
+  - fix check_backend_status in order to ignore unknown status
+  - fix service sequences_exhausted to take account of sequence's minvalue
+  - fix sequences_exhausted to take account of sequences only in the current db
+  - fix exclude option in backends_status service
+  - ignore startup and backup replication states in service streaming_delta
+  - avoid warning for -dev versions in pga_version service
+  - add a new uptime service
+  - replication_slots service handle wal files and pg_replslots files separately
+  - add ability to filter by application_name in longest_query and oldest_xact service
```
rjuju: do you mean oldest_idlexact?

check_pgactivity
```diff
@@ -3023,7 +3023,7 @@ sub check_commit_ratio {
     };
 
     if ( defined $args{'warning'} ) {
-        my $thresholds_re = qr/(rollbacks|rollback_rate|rollback_ratio)\s*=\s*(\d+)/i;
+        my $thresholds_re = qr/(rollbacks|rollback_rate|rollback_ratio)\s*=\s*([0-9]+\.?[0-9]*)/i;
```
rjuju: Since you can already specify both a number / rate / ratio, I'm not sure that this is required.

Also, decimals don't really make sense with `rollbacks` label.

anayrat: I disagree, number and rate are absolute value whereas ratio is relative. When traffic grows number and rate goes up whereas ratio can remain stable.

CHANGELOG.md
anayrat: Right! I'll fix. Thanks!

CHANGELOG.md
anayrat: Good catch! I'll fix. Thanks!

anayrat:
Hi,
Should I revert 39fffce to not block 2.4 release?
Thanks!

Krysztophe:
Is the 2.4 supposed to be v11 compatible? 
If yes, #196 is missing and other things still missing (see #192)
Anyway, as bugs, there is #194 (or I've misunderstood something)

anayrat:
No, we are far away to have chek_pga compatible with v11. I propose to release to 2.4 to fix several issues (look in changelog).

gleu:
How about someone merge this, and we start working on a fracking late v11 update?

Issue #155, opened 2018-01-10
backend_status returns garbage


Krysztophe:
Since #147, backend_status may return garbage if the queries in pg_stat_activity contains some binary, especially x03 or x30 
List of status may be corrupted and OPM rejects the perfdata.

See 6607a7a915612f038074aa55f6bfff4c6c0fb607 to mitigate the problem.

A translate in the query text may be necessary to avoid this.
Option --record-separator-zero  is unavailable before 9.2

@frost242 : remember 8565

frost242:
This issue can be fixed by filtering out the unwanted characters from the query field with the translate/chr function as here : 
```
            SELECT CASE
                WHEN s.waiting = 't' THEN 'waiting for lock'
                WHEN s.query = '<insufficient privilege>'
                    THEN 'insufficient privilege'
                WHEN s.state IS NULL THEN 'undefined'
                ELSE s.state
              END,
              extract('epoch' FROM
                date_trunc('milliseconds', current_timestamp-s.xact_start)
              ), current_setting('max_connections'),
              **translate(translate(s.query, chr(3), ''), chr(30), '') as query**
            FROM pg_stat_activity AS s
              JOIN pg_database d ON d.oid=s.datid
            WHERE d.datallowconn
```

@ioguix thoughts ?

ioguix:
This would have to be done in all queries returning user content other than numbers, dates, boolean etc, ie. bytea, text, xml, json, blablabla.

There's no so much places in check_pga, but anyway, it will be easy to forget this rule in future devs.

So -1. We should rather fix the way `query()` is parsing results.

frost242:
The most problematic part is the query or current_query column from pg_stat_activity. We don't rely on any query returning user content expect in backends_status. Imho, we can trust any data that are controled by a DBA.

Anyway, I don't see how we could fix the parsing in query() as the query column can hold any character that we may use in.
One possible way to fix this is to make two code paths: one for pg<9.2 and one for pg9.2+. We can't do much for the first case, but we can use the zero character as record separator for newer releases. We can also use --field-separator-zero but then we'll have to know how much columns the query should return, in order to form the recordset correctly. Do you feel this is a better way to handle the issue ?

ioguix:
> Remenber, @frost242 : we add once chr(30) appearing in pg_stat_activity.queries from one of our beloved customers.
>
[...]

PR #204, opened 2018-09-24
Mitigate gh issue #155

ioguix:
Hello,

Here is a patch which is imho a better step toward a clean solution for #155.

When a x30 or x03 appears in a result, it change the number of rows in a way or an other. This patch just checks that all rows in a result set has the same number of column, raisins an UNKNOWN with the error message if they diverge.

This solution is not perfect, it could be tricked by a value having the proper serie of x30 and x03, but I believe it's a cleaner way to mitigate the problem right now.


Krysztophe:
Everything is fine on the customer's server where it is currently in place since 10 days (no errors raised though). Anyway, this should be part of 2.4.

PR #207, opened 2018-11-07
Allow psql connections without any connection parameters given

tobixen:
Re issue #94 and earlier pull request #96, a new attempt on a pull request, rebased on the master, and with improved inline documentation.

Prior to this commit, if no connection parameters were given, one
would get errors like:

   FATAL: you must give only one host with service "XXX"

At the other hand, if at least one connection parameter was given,
any connection parameter omitted would cause other parameters to
be set to default.

After this commit, if no connection parameters are given, psql will be started up without any parameters, this will typically cause it to try to connect to the database using the default unix socket file in the default unix socket dir, typically /var/run/postgresql/.s.PGSQL.5432 (but may vary depending on distribution/compile-time options), with the same database username as unix username and same database name as unix username.

For backward compatibility (and to keep a minimal changeset), if any connection parameters are given, any omitted connection parameters will receive the default values as defined in the script.  This is arguably inconsistent, but I still believe this is the best option.

If you do believe consistency is more important than backward compatibility, I will rewrite the host parsing and psql connection so that the host will be omitted rather than defaulted to "localhost" when calling on psql.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -270,7 +270,15 @@ list.
 
 =item B<-h>, B<--host> HOST
 
-Database server host or socket directory (default: "localhost").
+Note: if neither host, username, port or database is given, psql will
+be invoked without any of those parameters passed over.  psql will
+then typically try to connect to the default unix socket (typically
+/var/run/postgresql/.s.PGSQL.5432) using the unix username as the
+database username and database, typically using the ident
+authentication.  If one or more of the above mentioned parameters are
+given but not all, the defaults will apply to the rest.
+
+Database server host or socket directory (default: "localhost")
 
 =item B<-U>, B<--username> ROLE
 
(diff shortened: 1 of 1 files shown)


ioguix:
Hello,

Your patch as is as some minor issues:
* documentation is not fully fixed (eg. header of parse_host)
* chosen name for the connection is not accurate. It doesn't feel right to use "localhost" or "local" or a combinaison or both. "binary defaults" seems more appropriate to me.
* if no connection arguments are given, I'm in favor to keep 100% of the binary defaults behavior where possible. That means it should relies on libpq env variable when they exists.

Moreover, when using one of `--dbname`, `--port`, `--username` or `--host` args, defaults are either the existing environment variables or the ones set in check_pga as a fallback. I'll clarify this in the doc.

As we plan to release something tomorrow, I'm currently working on this patch myself. I hope I'll commit something today if I am able to finish it on time.

Cross fingers, stay tune, and sorry for the lag :)

ioguix:
I gathered your commits in my own PR to keep authors and history. I'm closing this one. See #213

PR #210, opened 2018-12-17
handle archives timeline part as hexadecimal

pgstef:
With archives like "0000000B000000000000005B" in the directory, we get the following error:
```
Argument "0000000B" isn't numeric in addition (+) at ./check_pgactivity line 1543.
```

Because it was handled as decimal, not hexadecimal.


ioguix:
Shame on me and thanks to you.

PR #216, opened 2019-01-29
Fix archiver check on standby

### LINKED ISSUE #200: Archiver check should not return code 3 "UNKNOWN" on standby
Hi,

I am using the archiver check indifferently on a primary and on a standby (I don't want to reconfigure my probes if I have to migrate the primary).

Currently, this check is returning code 3 "UNKNOWN" as following:

```sh
# /usr/bin/sudo -u postgres /usr/local/nagios/plugins/check_pgactivity -h /var/run/postgresql --status-file /tmp/check_pgactivity.data -F nagios -s archiver
CHECK_PGACTIVITY UNKNOWN: Query fail !
psql:/tmp/check_pga-xFQPvg30:16: ERROR:  recovery is in progress
HINT:  pg_walfile_name() cannot be executed during recovery.
# echo $?
3
```

I think this check should return 0 "OK" on a standby instead of 3 "UNKNOWN".

ioguix:
Usage of 'pg_walfile_name()' is forbidden on standby. The fix makes the query
even more difficult to read, but we now parse the current LSN to compute the
number of WAL produced so far and compare with last archived one.

Fix #200

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -1761,13 +1761,20 @@ sub check_archiver {
 
     }
     else {
-        # Version 10.0 and higher: use pg_stat_archiver as the monitoring
-        # user may not be super-user.
+        # Version 10 and higher: use pg_stat_archiver
+        # as the monitoring user may not be super-user.
+        # FIXME: on a slave with archive_mode=always:
+        #  1) fails while parsing an .history file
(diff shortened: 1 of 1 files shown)


xals:
Works great, thank you very much!

PR #237, opened 2019-09-17
Fix broken archiver service

### LINKED ISSUE #236: service archiver does not warn for v10+
For v10 and after, the archiver service only check the number of waiting WAL. It does not report when the archiver is currently failing. This is a bug. This bug is especially bad if there's no threshold set up...

ioguix:
The archiver check was not working with pgsql 10 and after. It was
not complaining when the archiver was failing, neither reporting
the number of waiting WAL to archive.

This fix #236 on github

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -1776,28 +1776,52 @@ sub check_archiver {
         #  2) pg_last_wal_receive_lsn always returns zero if the slave is fed
         #     with pure log shipping (streaming is ok)
         my $query = q{
-        WITH s AS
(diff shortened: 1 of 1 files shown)


Krysztophe:
Sounds good to me.

As an example of result, here is an OPM graph on a v11 instance. First the archiver was broken (`/bin/false`), and then set back to normal. Then it was slowed down (`sleep 20` before `archive_xlog`), while some WALs were slowly and regularly added in the background, hence the odd shape.
archiver

@tilkow : I suppose this is what your customer wants: you can see a slow decrease of the delay, instead of a sudden drop like before (although we'll still often see that: archiving, once gone back to normal, flushed of the accumulated WALs rather quickly).

ioguix:
Good! Thank you for your tests and feedback @Krysztophe!

PR #239, opened 2019-11-18
Fix table bloat for v12: add last version from ioguix (v12 only)

### LINKED ISSUE #238: table_bloat : fails with PG 12
> psql:/tmp/check_pga-42bdwtPa:49: ERROR: column tbl.relhasoids does not exist LINE 34: + CASE WHEN tbl.relhasoids THEN 4 ELSE 0... ^

TODO : Update from the last version from the query: https://github.com/ioguix/pgsql-bloat-estimation/blob/master/table/table_bloat.sql

I'm just wondering if I create a whole new query for v12 (safe way) or if I update the current one (>9.0).

Krysztophe:
Fix #238 

For v12 only, use the last version from ioguix's table bloat query.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -6875,8 +6875,12 @@ sub check_table_bloat {
     my @dbexclude = @{ $args{'dbexclude'} };
     my $me        = 'POSTGRES_TABLE_BLOAT';
     my %queries   = (
-        # The following queries come straight from:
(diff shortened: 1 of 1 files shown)


ioguix:
Hi, I updated the query to use pg_statistic instead of pg_stats (as other queries does) for performances. I added some comments as well.

Thank for the PR.

PR #242, opened 2019-12-17
Add a json and json_strict output format.

rjuju:


check_pgactivity
```diff
@@ -1377,13 +1387,62 @@ sub nagios_strict_output ($$;$$$) {
...
+            'val' => $$_[1],
+            'unit' => $$_[2],
+            'warn' => $$_[3],
+            'crit' => $$_[4],
+            'min' => $$_[5],
+            'max' => $$_[6] }
+    } @perfdata;
+    $obj->{'perfdata'} = \%data;
+
+    print JSON::PP->new()->encode( $obj );
```
ioguix: Just use `encode_json $obj` here.

rjuju: Thanks, fixed!

rjuju:
Thanks @ioguix

PR #245, opened 2019-12-18
Replace NaN with "U" for strict outputs

### LINKED ISSUE #227: performance data contains Nagios-incompatible NaN values
We use check_pgactivity with Icinga, and it seems that performance data contains Nagios-incompliant `NaN` values. This results in lots of warnings about invalid performance data.
This became apparent for us while using the `longest_query` service.

Unless there is a more obvious fix for this that we don't know of, a possible solution for this would be to include in the `nagios` and/or `nagios_strict` output format options to replace `NaN` with `U` (according to https://nagios-plugins.org/doc/guidelines.html#AEN200)

>value, min and max in class [-0-9.]. Must all be the same UOM. value may be a literal "U" instead, this would indicate that the actual value couldn't be determined

Unfortunately I do not speak Perl much, so i can't submit a proper patch for this. It would be very nice if this could be fixed, however!

ioguix:
As of perfdata specs:
https://nagios-plugins.org/doc/guidelines.html#AEN200

Reported by Bas Gebbing on Github
Fix #227

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -1392,6 +1392,8 @@ sub nagios_strict_output ($$;$$$) {
 
     set_strict_units ( \@perfdata );
 
+    map { $$_[1] = 'U' if $$_[1] eq 'NaN' } @perfdata;
+
     return nagios_output( $rc, $ret, \@msg, \@perfdata, \@longmsg );
 }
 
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -1392,6 +1392,8 @@ sub nagios_strict_output ($$;$$$) {
 
     set_strict_units ( \@perfdata );
 
+    map { $$_[1] = 'U' if $$_[1] eq 'NaN' } @perfdata;
```
rjuju: How how instead changing `set_strict_units()` to say `set_strict_perfdata` that would be responsible of taking care of both units and NaN?

ioguix: You are right obviously. Fixed

PR #247, opened 2019-12-19
Rename ok() & other to avoid conflict with Test modules

### LINKED ISSUE #137: Subroutine ok() in conflict with the Test modules
The subroutine ok(), used to output an OK state conflicts with the Test modules that we may use in the future to implement non-regression tests.

This does not block anything until we use the Test::Harness infrastructure to test check_pgactivity.

ioguix:
Fix #137


ioguix:
thanks!

Issue #201, opened 2018-08-31
Monitor backend_xmin


anayrat:
Hi,

When you use hot_standby_feedback you may want to monitor xmin delta between primary and secondary to be alerted in case of long running transaction on secondary which cause bloat on primary.

Regards,

ioguix:
Can you elaborate here? 
What's the point as the same xact running on the primary would result in the same bloat? 
What would be the threshold criteria?
How would you set it?

I would rather add in `backends_status` the label `xact_age` (whatever its status) to check long running xact on both primary or secondary.

anayrat:
> Can you elaborate here?
> What's the point as the same xact running on the primary would result in the same bloat?


The issue is similar to have a long running xact on a primary : we can't vacuum tuple. For example if you have a long running pg_dump on a standby server, this could degrade primary performance: index containing old tuple, HOT chain not cleaned etc

I've seen write spikes on primary server when the xact is "freed", backends can prune record by themselves and when you have a big workload this could lead to production incident.

> What would be the threshold criteria?
> How would you set it?

Hard to say, I mostly depends on your workload. Maybe 100K transactions is enough?

> I would rather add in `backends_status` the label `xact_age` (whatever its status) to check long running xact on both primary or secondary.

That's a good idea but I think we should also check backend_xmin in pg_stat_replication from primary.

rjuju:
I agree that retained xmin horizon can be a real problem in some environment.

Counter proposal: a new check, say oldest_xmin, that would give the oldest age(xmin) for usual cause:
* single query
*  transaction (when xact_start != query_start)
* 2PC
* replication slot
* walsender

+ the overall oldest.  This way you can easily know if and why bloat is accumulating

ioguix:
@anayrat @rjuju: Good for me. It would requires some more specs now. And who want to do it now?

rjuju:
I'm volunteering to work on that.

ioguix:
done

PR #248, opened 2019-12-23
Add a new oldest_xmin check.

rjuju:
This implements the check mentioned in https://github.com/OPMDG/check_pgactivity/issues/201.

I'm not extremely sure of the pg_stat_activity part.  The problems I can see are:

- as soon as eg. a transaction retains some xid, active queries started after will also look like they're retaining the same amount of transactions.  However, if t hose queries continue to run after the initial transaction is finished, they'll be the source of retention.  So it's quite hard to really identify what actually retains the xmin horizons from an aggregated view.  The query is already complex enough, so I'm not in favor of adding extra steps to try to be smarter about that.

- For pg_stat_activity, I'm not exactly sure if both backend_xmin *and* backend_xid can both be earlier than the other.  Minimal testing showed than `coalesce(backend_xmin, backend_xid)` was enough, but I certainly didn't try a lot of scenario.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -426,6 +426,7 @@ my %args = (
     'port'                  => undef,
     'dbname'                => undef,
     'dbservice'             => undef,
+    'detailed'              => 0,
     'warning'               => undef,
     'critical'              => undef,
     'exclude'               => [],
@@ -5740,8 +5741,10 @@ sub check_oldest_xmin {
     my @msg_crit;
     my @msg_warn;
     my @hosts;
+    my $detailed;
     my $c_limit;
     my $w_limit;
+    my %oldest_xmin; # track oldest xmin and its kind for each database
     my %args      = %{ $_[0] };
     my $me        = 'POSTGRES_OLDEST_XMIN';
     my @dbinclude = @{ $args{'dbinclude'} };
@@ -5858,8 +5861,9 @@ sub check_oldest_xmin {
     ) unless $args{'warning'}  =~ m/^([0-9.]+)$/
         and  $args{'critical'} =~ m/^([0-9.]+)$/;
 
-    $c_limit = $args{'critical'};
-    $w_limit = $args{'warning'};
+    $detailed = $args{'detailed'};
+    $c_limit  = $args{'critical'};
+    $w_limit  = $args{'warning'};
 
     @hosts = @{ parse_hosts %args };
 
@@ -5879,10 +5883,23 @@ sub check_oldest_xmin {
 
         map { $_ = 'NaN' if $_ eq ''} @{$r}[2..3];
 
-        push @perfdata => (
-            ["$r->[0]_$r->[1]_age", $r->[2]],
-            ["$r->[0]_$r->[1]_xmin", $r->[3]]
-        );
+        if ($detailed) {
+            push @perfdata => (
+                ["$r->[0]_$r->[1]_age", $r->[2]],
+                ["$r->[0]_$r->[1]_xmin", $r->[3]]
+            );
+        }
+        else {
+            if ( exists $oldest_xmin{$r->[0]} ) {
+                $oldest_xmin{$r->[0]} = [ $r->[1], $r->[2] ]
+                    if $oldest_xmin{$r->[0]}[1] eq 'NaN'
+                    or $r->[2] > $oldest_xmin{$r->[0]}[1];
+            }
+            else {
+                $oldest_xmin{$r->[0]} = [ $r->[1], $r->[2] ];
+            }
+        }
+
         if (defined $c_limit) {
             if ($r->[2] ne 'NaN' and $r->[2] > $c_limit) {
                 push @msg_crit => "$r->[0]_$r->[1]_age";
@@ -5894,13 +5911,25 @@ sub check_oldest_xmin {
         }
     }
 
+    if (not $detailed) {
+        foreach my $k (keys %oldest_xmin) {
+            push @perfdata => (
+                ["${k}_age", $oldest_xmin{$k}[1]]
+            );
+
+            push @msg, "Oldest xmin in $k from ". $oldest_xmin{$k}[0]
+                if $oldest_xmin{$k}[1] ne 'NaN';
+        }
+    }
+
     return status_critical( $me, [
         'Critical: '. join(',', @msg_crit)
-        . (scalar @msg_warn? ' Warning: '. join(',', @msg_warn):'')
+        . (scalar @msg_warn? ' Warning: '. join(',', @msg_warn):''),
+        @msg
     ], \@perfdata ) if scalar @msg_crit;
 
     return status_warning( $me,
-        [ 'Warning: '. join(',', @msg_warn) ], \@perfdata
+        [ 'Warning: '. join(',', @msg_warn), @msg ], \@perfdata
     ) if scalar @msg_warn;
 
     return status_ok( $me, \@msg, \@perfdata );
@@ -8562,6 +8591,7 @@ GetOptions(
         'dbexclude=s',
         'dbinclude=s',
         'debug!',
+        'detailed!',
         'dump-status-file!',
         'dump-bin-file:s',
         'effective_cache_size=i',
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -165,6 +165,10 @@ my %services = (
         'sub'  => \&check_oldest_idlexact,
         'desc' => 'Check the oldest idle transaction.'
     },
+    'oldest_xmin' => {
+        'sub'  => \&check_oldest_xmin,
+        'desc' => 'Check the xmin horizon per distinct sources of xmin retention.'
+    },
     'longest_query' => {
         'sub'  => \&check_longest_query,
         'desc' => 'Check the longest running query.'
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -5600,6 +5604,167 @@ sub check_oldest_idlexact {
...
+    $w_limit = $args{'warning'};
+
+    @hosts = @{ parse_hosts %args };
+
+    pod2usage(
+        -message => 'FATAL: you must give only one host with service "oldest_xmin".',
+        -exitval => 127
+    ) if @hosts != 1;
+
+    is_compat $hosts[0], 'oldest_xmin', $PG_VERSION_84 or exit 1;
```
ioguix: I wonder what is the point to support 9.3 and before as we only check for prepared xact. There's already the `oldest_2pc` check that raise an alert for stalled prepared xact.

check_pgactivity
```diff
@@ -5600,6 +5604,167 @@ sub check_oldest_idlexact {
...
+            }
+
+            push @msg_warn => "$r->[0]_age"
+                if ($r->[1] ne 'NaN' and $r->[1] > $w_limit);
+        }
+    }
+
+    return status_critical( $me, [
+        'Critical: '. join(',', @msg_crit)
+        . (scalar @msg_warn? 'Warning: '. join(',', @msg_warn):'')
```
ioguix: lack a space betwwen the critical list and the `Warning` label?

check_pgactivity
```diff
@@ -154,6 +154,10 @@ my %services = (
         'sub'  => \&check_oldest_idlexact,
         'desc' => 'Check the oldest idle transaction.'
     },
+    'oldest_xmin' => {
+        'sub'  => \&check_oldest_xmin,
+        'desc' => 'Check the xmin horizon per distinct sources of xmin retention.'
```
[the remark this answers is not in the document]
rjuju: That sounds weird.  **for** distinct...?

check_pgactivity
```diff
@@ -5600,6 +5604,167 @@ sub check_oldest_idlexact {
...
+source doesn't retain any transaction, NaN is returned.
+For versions prior to 9.4, only C<2pc> source of xmin retention is available,
+so other sources won't appear in the perfdata.
+
+Critical and Warning thresholds only accept a raw number of transaction.
+
+Required privileges: an unprivileged role checks only its own queries;
+a pg_read_all_stats (10+) or superuser (<10) role is required to check
+pg_stat_replication.  2PC, pg_stat_activity, and replication slots don't
+require special privileges.
```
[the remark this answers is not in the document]
rjuju: Do you have a suggestion for a better wording?

check_pgactivity
```diff
@@ -5600,6 +5604,167 @@ sub check_oldest_idlexact {
...
+    my @msg_crit;
+    my @msg_warn;
+    my @hosts;
+    my $c_limit;
+    my $w_limit;
+    my %args        = %{ $_[0] };
+    my $me          = 'POSTGRES_OLDEST_XMIN';
+    my %queries     = (
+        # 8.4 is the first supported version as we rely on window functions to
+        # get the oldest xmin.  Only 2PC has transaction information available
```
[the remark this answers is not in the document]
rjuju: Probably yes, although I'm not sure it's worth the trouble.

check_pgactivity
```diff
@@ -5600,6 +5604,167 @@ sub check_oldest_idlexact {
...
+            ORDER BY age(xmin) DESC NULLS LAST
+          ) rownum, age(xmin) AS age, xmin
+          FROM raw
+          RIGHT JOIN (
+            SELECT           'query'
+            UNION ALL SELECT 'idle_xact'
+            UNION ALL SELECT 'active_xact'
+            UNION ALL SELECT '2pc'
+            UNION ALL SELECT 'repslot'
+            UNION ALL SELECT 'walsender'
```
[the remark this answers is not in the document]
rjuju: If you want.  Do you know since when this is supported?

check_pgactivity
rjuju: Because a 1m old 2pc holding 10 transactions is not the same as a 1min old 2pc holding 1M transactions, and if you have a non uniform xid generation it's likely to have random bloat issue that oldest_2pc won't be able to help without a lot of noise?

check_pgactivity
rjuju: Either I'm missing something or the problem exists for every other calls to status_critical.

check_pgactivity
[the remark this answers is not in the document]
ioguix: I'm sure as I did not check what privileges were required from various place. But as far as I understand your wording, what about:

> Required privileges: an unprivileged role can check xmin from 2PC, pg_stat_activity, and 
> replication. a pg_read_all_stats (10+) or superuser (<10) role is required to check
> pg_stat_replication. 

I don't get the difference between `an unprivileged role checks only its own queries`  and `pg_stat_activity don't require special privileges` though...

check_pgactivity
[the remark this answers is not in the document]
ioguix: Since this query is for 9.4 and after, do we care?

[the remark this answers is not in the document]
rjuju: oh right, indeed.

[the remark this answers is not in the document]
ioguix: For the completeness of my answer, this has been introduced in 8.2. See chapter https://www.postgresql.org/docs/8.2/release-8-2.html#AEN80530

check_pgactivity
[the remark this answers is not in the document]
rjuju: Actually after re-reading pg_stat_get_activity() I think that the "an unprivileged role checks only its own queries;" is a missed copy/paste and should be removed, thus this should be ok:

```
Required privileges: a pg_read_all_stats (10+) or superuser (<10) role is required to check
pg_stat_replication.  2PC, pg_stat_activity, and replication slots don't
require special privileges.
```

check_pgactivity
ioguix: Well, ok.

FYI, I can test all versions up to 8.1 where pg_prepared_xacts has been introduced. So you can go the full way down this abyssal version.

check_pgactivity
ioguix: Most of status_critical call I found are using multiple array entries whereas you are doing a string concatenation.

check_pgactivity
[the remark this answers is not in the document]
ioguix: It will be the same query for all version down to 8.1:

~~~
postgres@postgres=# SELECT version();
 PostgreSQL 8.1.23 on x86_64-unknown-linux-gnu, compiled by GCC gcc (Debian 8.3.0-6) 8.3.0

postgres@postgres=# \d pg_prepared_xacts 
 transaction | xid                      | 
 gid         | text                     | 
 prepared    | timestamp with time zone | 
 owner       | name                     | 
 database    | name                     | 
~~~

check_pgactivity
rjuju: I see.  Somehow I managed to copy/paste the only broken call: https://github.com/OPMDG/check_pgactivity/blob/master/check_pgactivity#L5170-L5174

check_pgactivity
rjuju: Thanks!

check_pgactivity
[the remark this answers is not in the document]
ioguix: as discussed, I still prefer __from__

check_pgactivity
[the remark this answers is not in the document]
ioguix: works for me.

check_pgactivity
ioguix: Oops.

Fixed: a09b8b823bd0529

Thanks for the report :)

rjuju:
As discussed off thread, I realized that I forgot to also handle the impacted databases.  I just pushed an additional commit (that should be squashed, after a rebase I'll take care of that once everything is ok) to fix both that and @ioguix comments.

check_pgactivity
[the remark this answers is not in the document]
ioguix: This still need to be fixed.

check_pgactivity
```diff
@@ -5600,6 +5604,186 @@ sub check_oldest_idlexact {
 }
 
 
+=item B<oldest_xmin> (8.4+)
+
+Check the xmin I<horizon> from distinct sources of xmin retention.
+
+Perfdata contains the oldest xmin and maximum age for the following source of
+xmin retention: C<query> (a currently executing query), C<active_xact> (an
+opened transaction currently executin a query), C<idle_xact> (an opened
```
ioguix: > currently executin**g**

check_pgactivity
```diff
@@ -5600,6 +5604,186 @@ sub check_oldest_idlexact {
...
+transaction being idle), C<2pc> (a pending prepared transaction), C<repslot> (a
+recpliation slot) and C<walwender> (a WAL sender replication process), for each
```
ioguix: > C<wal**s**ender>

check_pgactivity
```diff
@@ -5600,6 +5604,186 @@ sub check_oldest_idlexact {
...
+          ) f(kind) ON raw.kind = f.kind
+          -- we use this JOIN condition to make sure that we'll always have a
+          -- full record for all (connectable) databases
+          JOIN pg_database d ON d.datname = coalesce(raw.datname, d.datname)
+          WHERE d.datallowconn
+        )
+        SELECT kind, datname, age, xmin
+        FROM ordered
+        WHERE rownum = 1
+        }
```
ioguix: This query reports as many walsender than existing database, which is wrong.

check_pgactivity
rjuju: will fix

check_pgactivity
rjuju: will fix

check_pgactivity
rjuju: This returns as many physical walsender as connectable database, which is exactly what is expected.  It simplifies knowing what is the xmin horizon on a given database without checking the possible "global" limits.

ioguix: physical walsender are global, they are not related to one single database. So I don't understand your answer. It's still confusing

check_pgactivity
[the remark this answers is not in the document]
ioguix: whatever, this is not very important.

check_pgactivity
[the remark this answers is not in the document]
ioguix: well, ok, I give up on that as well.

check_pgactivity
```diff
@@ -5600,6 +5604,186 @@ sub check_oldest_idlexact {
...
+
+    is_compat $hosts[0], 'oldest_xmin', $PG_VERSION_84 or exit 1;
+
+    @rs = @{ query_ver( $hosts[0], %queries ) };
+
+    REC_LOOP: foreach my $r (@rs) {
+        map { $_ = 'NaN' if $_ eq ''} @{$r}[2..3];
+        push @perfdata => (
+            ["$r->[0]_$r->[1]_age", $r->[2]],
+            ["$r->[0]_$r->[1]_xmin", $r->[3]]
```
ioguix, line 5761: I still think reporting 6 kinds * 2 metrics * num_db  is very a lot of metrics, but anyway. I can give up on this as far as we can move on this PR. We'll be able to discuss and find solutions later if it is a real pb raising from the field.

However, using `"$r->[1]_$r->[0]_age` (`$db_$kind_age|xmin`) reads better to me. Eg. `dbname_walsender_age` instead of `walsender_dbname_age`. "age" is related to "kind". Moreover, it is easier to sort by dbname when checking metrics by hands.

rjuju: The only way to reduce then number of metrics is to report physical slots only once, as this is the only global metric.  That probably won't save much, but I won't object to that change.  Special care will have to be taken for the documentation to help users determining what is the xmin horizon on a specific database, as this won't be as obvious as before.

check_pgactivity
ioguix: I understood why it has been done this way after discussing. Closing this conversation

check_pgactivity
ioguix: As discussed, I'm ok with this.

ioguix (approved):

ioguix:
Hey @rjuju,

I've been reviewing and testing. I closed all our active conversations as we finished discussing all of them.

I fixed a bug in the query were some `NaN` perfdata (`idle_xact` or `active_xact` or `query`) were not shown for database when they were set in other databases.

Please, do some test/review and fixup/merge when you are ok.

ioguix:
BTW, with only 4 connectable db (including `template1` and `postgres`), I gathered 48 perfdatas. 12 metrics per database.

Maybe we can add a `--summary` arg or equivalent to only provide in perfdata the oldest xmin/age per database, and report in long messages what is the oldest horizon for each database?

frost242:
+1 for a `--summary` option.

rjuju:
Thanks @ioguix 
I'm fine with a `--summary` option.  However, just to be clear I don't expect anyone using that check for all databases nor for a long time.  It's really a diagnosis check to help track down bloat issue happening at some random interval, but the rest of the time you don't really need to monitor it.

I'll review the changes soon.

ioguix:
As discussed on IRC, I added a few commits:

* thresholds are now optionnal
* the check is now terse by default: one perfdata (oldest age) per database plus one info per database (what is the kind of the oldest xmin) in messages if there's some data to show.
* new arg `--detailed` throw detailed perfdatas (12) for each database

Please, review.

++

rjuju:
I'm fine with the new behavior.  About the patch, I think we should document the new `--detailed` option along with the other ones. 

I also see that you don't check for compatibility with this option.  Is that intended?  I'm not sure if anyone will blindly add a `--detailed` everywhere just in case so that's may not be a huge problem, but until now our current approach is to explicitly check for every single option compatibility.

ioguix:
>  About the patch, I think we should document the new --detailed option along with the other ones.

Well, a lots of specifics argument are not documented with other, normal, ones. Specifics ones are only documented in their related service. Eg.: `--query`, `--path`, `--type`, `--suffix`, `--uid`, GUCs, `--unarchiver`, etc.

> I also see that you don't check for compatibility with this option. Is that intended?

Fixed in last commit.

PR #250, opened 2020-03-24
Fix backend count for PostgreSQL version 10 and a typo in 12.

blogh:
- query PG_VERSION_100: background workers shouldnt count towards
  client_backends. With this patch 'backend_type' is now restricted
  to 'client backend' only since walsenders are accounted for using
  pg_stat_replication.

- query PG_VERSION_120: the modification is the same here. Except that
  the result of the query was correct because the backend_type
  'background worker' doesn't exist anymore, it was replaced by 'logical
  replication worker' and 'parallel worker' in version 11.


rjuju:
Good catch!  Thanks a lot, merged.

Issue #253, opened 2020-06-23
Annoying dependency to JSON::PP


yhuelf:
Commit 752aea91 adds a dependency to JSON::PP, but this Perl module isn't always installed by default on database servers.

Would it be possible to dynamically import this module at runtime only when JSON output is used?

rjuju:
@yhuelf could you try with latest master and confirm that @frost242 patch is fixing the problem for you?

yhuelf:
Thanks! It works well :-)

frost242:
Thanks for the report.

PR #254, opened 2020-06-23
Fix check_pga to run on prehistoric Perl installation

frost242:
In dark ancient ages, Perl didn't provide JSON::PP by default. This PR fixes issue #253.

Regards


rjuju:
Thanks for taking care of this!

Issue #231, opened 2019-05-15
v12: use pg_ls_tmpdir() in check_temp_files?


ioguix:
A new function `pg_ls_tmpdir()` appear in v12.

frost242:
Done in commit  79b0d0d

PR #255, opened 2020-06-24
Use pg_ls_tmpdir() in check_temp_files

frost242:
PostgreSQL 12 introduced function pg_ls_tmpdir to list live temp files
in a given tablespace. This function requires the pg_monitor privilege

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7420,7 +7420,7 @@ that it cannot access, nor live temp files
  v11: an unprivileged role is possible but must be granted EXECUTE
 on functions pg_ls_dir(text), pg_read_file(text), pg_stat_file(text);
 the same restrictions than on v10 will still apply
- v12: a user with the role pg_monitor suffices.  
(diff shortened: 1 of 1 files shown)


frost242:
As per issue #231

check_pgactivity
```diff
@@ -7420,6 +7420,8 @@ that it cannot access, nor live temp files
  v11: an unprivileged role is possible but must be granted EXECUTE
 on functions pg_ls_dir(text), pg_read_file(text), pg_stat_file(text);
 the same restrictions than on v10 will still apply
+ v12: a user with the role pg_monitor suffices. unprivileged role is not
```
rjuju, line 7423: I'd personally drop the "unprivileged role is not supported anymore", as it actually wasn't the case before.

frost242: Done in commit 13324c7

frost242: Wording improved as per commit d3f3cda

PR #258, opened 2020-07-23
PG 13 is coming

Krysztophe:
This is the whole result of my first regression tests with PG13 beta2.


Krysztophe:
Well, it was short and informative, but I've changed it.

rjuju:
Thanks.

PR #265, opened 2020-10-07
do not error on missing file in check_temp_files for version 10+

madtibo:
In the check_temp_files, since we are first getting file list before getting their size, some files could get missing in the meantime. The check then fails with the following error:
```
CHECK_PGACTIVITY UNKNOWN: Query failed !
psql:/tmp/check_pga-pBGcyGwB:29: ERROR:  could not stat file base/pgsql_tmp/pgsql_tmp25560.109: No such file or directory
```

Update the SQL query for versions 10+ to return a size of 0 for the missing files.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7572,7 +7572,7 @@ sub check_temp_files {
         # PostgreSQL 10
         $PG_VERSION_100 => q{
           SELECT 'live', agg.spcname, count(agg.tmpfile),
-                 SUM((pg_stat_file(agg.dir||'/'||agg.tmpfile)).size) AS SIZE
+                 SUM(COALESCE((pg_stat_file(agg.dir||'/'||agg.tmpfile, true)).size, 0)) AS SIZE
             FROM ( SELECT ls.oid, ls.spcname AS spcname,
                           ls.dir||'/'||ls.sub AS dir,
                           pg_ls_dir(ls.dir||'/'||ls.sub) AS tmpfile
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7421,7 +7421,7 @@ Required privileges:
  v10: an unprivileged role is possible but it will not monitor databases
 that it cannot access, nor live temp files
  v11: an unprivileged role is possible but must be granted EXECUTE
-on functions pg_ls_dir(text), pg_read_file(text), pg_stat_file(text);
+on functions pg_ls_dir(text), pg_read_file(text), pg_stat_file(text, boolean);
 the same restrictions than on v10 will still apply
  v12+: a role with pg_monitor privilege.
 


frost242:
Hello,
Can you also update the documentation regarding the privilege required to run this check ?

madtibo:
I do not think this patch is changing anything about privileges.  
Furthermore, it seems this information is already present:

```
=item B<temp_files> (8.1+)

(...)

Required privileges:
 <10: superuser
 v10: an unprivileged role is possible but it will not monitor databases
that it cannot access, nor live temp files
 v11: an unprivileged role is possible but must be granted EXECUTE
on functions pg_ls_dir(text), pg_read_file(text), pg_stat_file(text);
the same restrictions than on v10 will still apply
 v12+: a role with pg_monitor privilege.
```
No ?

frost242:
Oh look, two functions with the same name ! I guessed that you knew that...

```
postgres=# \df pg_stat_file
Liste des fonctions
-[ RECORD 1 ]------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Schéma                         | pg_catalog
Nom                            | pg_stat_file
Type de données du résultat    | record
Type de données des paramètres | filename text, missing_ok boolean, OUT size bigint, OUT access timestamp with time zone, OUT modification timestamp with time zone, OUT change timestamp with time zone, OUT creation timestamp with time zone, OUT isdir boolean
Type                           | func
-[ RECORD 2 ]------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Schéma                         | pg_catalog
Nom                            | pg_stat_file
Type de données du résultat    | record
Type de données des paramètres | filename text, OUT size bigint, OUT access timestamp with time zone, OUT modification timestamp with time zone, OUT change timestamp with time zone, OUT creation timestamp with time zone, OUT isdir boolean
Type                           | func
```

madtibo:
There is no need to be sarcastic!  
If you know exactly what need to be updated, you could have pointed it in your first message.

So, no, I was not aware about these two functions, and I do not feel bad about it.  
There are many many things that I do not know about PostgreSQL and I am always happy to learn.

I will look into it to update the documentation.

Thank you in advance for your goodwill and kindness.

frost242:
Thanks for taking care. Apologies for the sarcasm, it was not welcome.

madtibo:
I added the boolean parameter to the pg_stat_file for v11.

All the listed functions exists in v10.  
I was wondering why the section about functions privileges was only present for v11 and not v10?

frost242:
Thanks a lot.
The section concerns only v11 and not v10 because in v10 you have to be superuser to execute pg_stat_file : 
```
	if (!superuser())
		ereport(ERROR,
				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
				 (errmsg("must be superuser to get file information"))));
```

See : https://github.com/postgres/postgres/blob/REL_10_STABLE/src/backend/utils/adt/genfile.c#L298

This check is no more present starting from version 11 : https://github.com/postgres/postgres/blob/REL_11_STABLE/src/backend/utils/adt/genfile.c#L383

Thanks, I merge your change, good point about using this function btw.

madtibo:
Thank you very much for the pg_stat_file info. I now see the difference in the documentation.

Bonne fin de journée !

Issue #270, opened 2020-10-30
psql 8.3 and before doesn't accept -w


ioguix:
This makes check_pga currently not compatible with such versions of psql.

Krysztophe:
As discussed in private:

- at least worth a doc patch, to tell the users to use another client
- we cannot know if ''trust'' is used, or if the service file contains the password, so it is complicated to add the ''-w'' dynamically
- this limit should not be a problem for many people nowadays
- testing psql at each query is overkill
- a ''--very_old_psql_client'' option to remove the ''-w'' shoud be enough + doc

frost242:
Thanks for the patch

ioguix:
In fact, this bug has been introduced in de404d9ca13e282 after short discussion in #229.

I wonder if we should revert the patch for this release and postpone in a better shape for next release?

rjuju:
Oops.  I guess we could cache the psql version in the storable file, but I'm not sure of how we should clear that cache.

Is it that common to have that old psql clients?  I'm not opposed to revert the patch to avoid postponing the next release.

Krysztophe:
I would keep the patch. Some people may already use it. There is a workaround (a less prehistoric client). If somebody really complains, add a ''--psql-is-prehistoric'' option. Or drop 8.3 and sooner support.

rjuju:
I'm fine with an extra option too.  I'm assuming you meant 8.3 and older?  Note that it's only the client side, server side shouldn't be a problem.

ioguix:
> Is it that common to have that old psql clients?

No, I don't think so. We have nothing that old on our side. But I have absolutely no clue outside there in the wild.

There might be another solution. Refactor the way we query the instance. This is a long overdue project in my opinion, to eg avoid one connection per query, detect psql version, parse correctly the return value (think #155), etc. Maybe next release might be a good time to fork a psql process and play with its stdin/stdout. We could then deal with psql version and interactive password for older ones to fail the connection quickly?

Krysztophe:
@rjuju : <8.3 , of course.  

@ioguix : We want to release asap, so let's keep it this way now.  8.3 is too old to really support, we have a good work around, and a 2-characters emergency patch.  You can do whatever you want in 2.6.

ioguix:
> You can do whatever you want in 2.6.

That's what I wrote.

The release process is on-going: #274

PR #271, opened 2020-10-30
Minimal version 7.4, and psql client 8.3

Krysztophe:
See #270 

Added a compatiblity section at the top: psql client must be 8.3 and the server at least 7.4. It was not written elsewhere. As some services are flagged "all versions", the server minimal version could be discussed.


frost242:
Ok for me, I do the merge, thanks !

PR #276, opened 2020-11-13
WIP: add TAP Tests

ioguix:
This is currently a work in progress branch adding TAP tests.

I am currently trying to create various tests examples so other can write some more based on existing one. Look at `t/README` and code comments about how to run them.

If you want to dig, the `streaming_delta` tests are the most complex presently. Some of them are wrapping `query()` sub to mock the real result and test `check_pga` reactions. Look at the lag-related checks.

This is still in WIP as I hadn't tests on other release than v13 presently.

This branch is subject to be rebased/history rewrote by push --force without notice :)

Any comment/advice welcome.


rjuju (commented):
Should the upstream modules be in a specific directory, for clarity?  It may also be good to have a script, makefile or some template to be able to fetch the latest upstream modules automatically, from a customizable directory.

rjuju (approved):

t/lib/pgNode.pm
```diff
@@ -0,0 +1,171 @@
...
+    my $stb = shift;
+
+    $self->{'node'}->wait_for_catchup($stb->{'node'}, @_);
+}
+
+=pod
+
+=head1 METHODS
+
+Bellow the changes and new methods implemented in this facet.
```
rjuju, line 84: s/bellow/below/

t/lib/pgNode.pm
```diff
@@ -0,0 +1,171 @@
...
+    my $wal          = shift;
+    my $max_attempts = 300; # 30s
+    my $walfile      = $self->archive_dir() ."/$wal";
+
+    print "# waiting for archive $walfile\n";
+
+    while ($max_attempts and not -f $walfile) {
+        $max_attempts--;
+        # Wait 0.1 second before retrying.
+        usleep(100_000);
```
rjuju, line 145: should we sleep a bit longer and try less times?  For now those tests will probably be run on dev's machine, but we might eventually rely on external CI and we probably won't be able to assume that those will have as much ressources.  In that case It may also be a good idea to test first and then sleep if needed.

Should also the sleep time be declared next to max_attempts?  It would make it clearer and more consistent.

t/lib/pgNode.pm
```diff
@@ -0,0 +1,171 @@
...
+
+Return the old segment filename.
+
+=cut
+
+sub switch_wal {
+    my $self = shift;
+    my $result;
+
+    if ($self->version > '9.6') {
```
rjuju, line 111: Does this still use the overloaded <=> and cmp operators (from PostgresVersion.pm)?

rjuju (changes requested):

t/01-archive_folder.t
```diff
@@ -0,0 +1,162 @@
...
+$node->switch_wal;
+$node->psql('template1', 'insert into t select generate_series(20001,30000) i');
+$wal = $node->switch_wal;
+$node->wait_for_archive($wal);
+
+### Begin of tests ###
+
+# simple success check
+$node->command_checks_all( [
+    './check_pgactivity', '--service'  => 'archive_folder',
```
rjuju, line 38: Note that this requires the script to be executable, which I think is not the case right now.  Is it?

I'm in favor of changing the mode bits rather than putting `perl ./check_pgactivity`  everywhere.

rjuju (approved):

t/lib/Mocker/Streaming.pm
```diff
@@ -0,0 +1,47 @@
...
+
+    # install wrapper around query sub to capture and modify the result
+    $main::{'query'} = sub {
+        my $res;
+
+        $res = $main::{'query_orig'}->(@_);
+
+        return $res unless $_[1] =~ m/FROM pg_stat_replication/;
+
+        # mock 1MB of write delta.
```
rjuju, line 32: I'm fine with this approach (especially since I don't have a better idea), but we should document that we're *not* testing that the queries are actually correct.  For instance, if someone swaps the write and flush delta in the query the tests won't be impacted.

t/lib/Mocker/Streaming.pm
```diff
@@ -0,0 +1,47 @@
...
+        #      2MB of flush delta.
+        #      3MB of replay delta.
+        # We don't mind the total WAL size (the X part of XXX/YYYYY) as the
+        # test stay far bellow it.
```
rjuju, line 36: s/bellow/below/

rjuju (changes requested):

t/README
```diff
@@ -0,0 +1,128 @@
+This folder contains all the tests related to `check_pgactivity`.
+
+# Environment setting
+
+The test are written using well known and used `Test::More` perl module.
+Bellow are instructions to setup basic environment to run them.
```
rjuju, line 6: s/bellow/below/

rjuju (changes requested):
Very nice work, thanks a lot!

I just have some minor comments:

* I think that we should emit the version we actually found, to avoid any problem when you think you tested all the major versions but actually didn't.  That's especially useful if you plan to run the tests with a loop and some custom tooling that dynamically change your env variable dynamically
* should we add some kind of script / makefile to run some basic requirements tests and give you friendlier info if you're missing part of the stuff needed (prove, IPC::Run and so on)?

t/lib/pgNode.pm
ioguix: Yes. See second table there: https://perldoc.perl.org/overload#Magic-Autogeneration

t/lib/pgNode.pm
ioguix: next patch declare the sleep time next to max_attempts. As discussed together, I kept the 0.1s sleep period.

t/01-archive_folder.t
ioguix: This has been committed in a1dfb35fce499e9147df50bc56d1c564caf7b2de

t/lib/Mocker/Streaming.pm
ioguix: fixed

t/README
ioguix: fixed

t/lib/pgNode.pm
ioguix: fixed

ioguix:
> Should the upstream modules be in a specific directory, for clarity? 

This would force to either change the module namespace or add another lib path to each test file. Eg.:
  - `use TestLib` becomes `use Upstream::Testlib`
  - or add `use lib 't/lib/Upstream'`

I don't want to use a different namespace for module, I suspect if might break some code relying on current module names.
I prefer to keep the test file boilerplate as simple and common as possible. If we should add `use lib 't/lib/Upstream` in some case

>It may also be good to have a script, makefile or some template to be able to fetch the latest upstream modules automatically, from a customizable directory.

Upstream is currently a moving part. Let's wait for the PostgresNodeNg to land in core to decide what to do?

ioguix:
> I think that we should emit the version we actually found, to avoid any problem when you think you tested all the major versions but actually didn't

Each test already show the version used as comment. You need to be verbose when using `prove` Eg.:

~~~
$ prove -v t/01-streaming_delta.t 
t/01-streaming_delta.t .. 
1..121
# Node "prim" uses version: 11.10
# Node "sec1" uses version: 11.10
# Node "sec2" uses version: 11.10
# testing on version 11.10
# primary started
[...]
~~~

Running tests by hands:

~~~
$ perl t/01-backends.t 
1..34
# Node "prod" uses version: 11.10
ok 1 - [...]
~~~

t/lib/Mocker/Streaming.pm
ioguix: Fixed. I added some explanations and warning in comments.

ioguix:
> should we add some kind of script / makefile to run some basic requirements tests and give you friendlier info if you're missing part of the stuff needed (prove, IPC::Run and so on)?

I'm not convinced, the requirements are very low and documented in `t/README`. Did you had any trouble finding what to install to run tests? Should I update the doc with some more directives to help any new comer?

Same comment and questions for running tests themselves. Do the doc lack some more info?

Issue #249, opened 2020-03-11
Lack of maintenance not detected


rjuju:
Looking at `check_last_maintenance`, I see that the code fails to properly account for tables that were never vacuumed or analyzed, which can lead to a "OK" result if that's the state of all the tables

To be more accurate, there's a coalesce to `NaN`, but the verification is done using `-inf`
(see https://github.com/OPMDG/check_pgactivity/blob/master/check_pgactivity#L4547 and https://github.com/OPMDG/check_pgactivity/blob/master/check_pgactivity#L4612)

frost242:
@rjuju the MIN() aggregate is also a problem to this issue. It does not take account of NULL values, so the only way you'll get a NaN is when there was no maintenance at all on the database.

frost242:
Anyone has a chance to take a look at my PR ?

Krysztophe:
Hi,

I've tested it, and it seems okay. Once a single simple unvacuumed table is added, I get a CRITICAL:
`POSTGRES_LAST_ANALYZE CRITICAL: pogo: -Infinity | banque=12311.288849s;86400;432000 christ=NaNs;86400;436400;432000 icingaweb=12346s;86400;432000 owncloud=12294.675267s;86400;432000 pgbench=12294.198117s;86400;432000 pogo=-Infinitys;;86400;432000`

The critical disapears after a VACUUM resp. ANALYZE. 
The unvacuumed/unanalyzed table raises the critical when other vacuumed tables exist.
A database without any table does not raise a warning (the query gives NaN).

I'm just wondering if users will like the change in behaviour. If you add a single table for any reason, you immediately get a critical. This is not really the same thing as a autovacuum that is lacking behind. On the other side, you do have unvacuumed tables.

So that's okay for me.

frost242:
Thanks, merged

ioguix:
According to #256, this is issue fixed

Krysztophe:
Reopening:  

with the new code,
when there are many temp tables, or the application like to create many tables, the result becomes CRITICAL / -Infinity because of all the empty last_analyze/last_autoanalyze. 

We can add `relpersistence = 't' ` for the temp tables, but it will not be enough. 

Adding `--exclude` may be another options.

But a very small table too small to have ever been analyzed will still trigger a CRITICAL.

(Same problem for vacuum, of course)

ioguix:
> We can add `relpersistence = 't'` for the temp tables, 

[...]

PR #278, opened 2020-11-18
Ignore temp tables in maintenance services

ioguix:
related to #249


Krysztophe:
Tested, temp tables are now ignored.

PR #285, opened 2021-03-23
Allow to set username and dbname when using Socket connexion

aaujon:
Allow to connect to pgsql with socket setting user and dbname.
Now, if you set host or port parameter it will use tcp connexion but without it will use socket connexion the same way pgsql does.
I also did some double assignment cleanup.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -893,10 +893,9 @@ sub parse_hosts(\%) {
     # Add as many hosts than necessary depending on given parameters
     # host/port/db/user.
     # Any missing parameter will be set to its default value.
+    # if you set host or port, it will user tcp connexion.
     if (defined $args{'host'}
-        or defined $args{'username'}
         or defined $args{'port'}
-        or defined $args{'dbname'}
     ) {
         $args{'host'} = $ENV{'PGHOST'} || 'localhost'
             unless defined $args{'host'};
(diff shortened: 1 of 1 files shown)


ioguix:
Hi,

Thank you for your patch.

Unfortunately, I'm afraid this could harm a lot of existing setups relying on this current default, consciously or not.

I kind of remember this default is set to `localhost` to be compatible with microsoft OS, when they were not supporting non-networking sockets.

> Allow to connect to pgsql with socket setting user and dbname

This is still possible, but admittedly you have to give more arguments than with psql. Indeed, the `-h` arg is not for tcp-only remote host, you can use `-h /path/to/socket/folder` if you want to connect through unix socket:

~~~console
$ perl -l check_pgactivity -s connection -h /tmp -p 15433
POSTGRES_CONNECTION OK: Connection successful at 2021-03-26 16:59:39.374157+01, on PostgreSQL 13.1 on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit

$ ls -l /tmp/.s.PGSQL.15433
srwxrwxrwx 1 ioguix ioguix 0 mars  26 16:36 /tmp/.s.PGSQL.15433
~~~

aaujon:
Thanks for your reply, I didn't notice that -h could be used for a socket. I think that solve my problem.
I agree with you about the fact it could break existings setups. Too bad it can't use the same default as pgsql.

PR #295, opened 2021-07-16
last_vacuum / last_analyze  s/min/max

yhuelf:
we want the oldest maintenance, so the max of the difference with
current_timestamp.


ioguix:
Good catch, thank you!

PR #303, opened 2022-02-09
Improve locking around the status file

### LINKED ISSUE #287: File «${storage}» not recognized as a check_pgactivity status file.
Hi,

I'm using nagios to check a postgres DB.
Sometime /var/local/pgactivity/pgactivity.data become empty (0byte) which forbid any future calls to be executed until I delete this file.
I guess there is a race condition somewhere.

I'm not sure it fixes my issue but in save method : The exclusive lock should be acquired before lock_retrieve line 567 and released after lock_store line 571. I think with the current code 2 processes can  read the file at the same moment and the last one will override the first one write.

The problem occurs everyweek or so, I can provide more information if you need.

Thanks,

ioguix:
Our previous locking policy could lead to some dead locks and status
file truncation.

This patch tries to avoid dead locks by covering the whole content
of "save()" under the same exclusive lock. Similary, the whole
content of "load()" is now covered by only one shared lock.

Plus, to avoid any futur race condition between the "open()" and
"flock()" calls, we are now using an external lock file to hold our
advisory lock.

Fix #287


check_pgactivity
```diff
@@ -552,24 +549,32 @@ sub find_hostkey($) {
...
-    my $hostkey = find_hostkey($host);
+    my $host     = shift;
+    my $name     = shift;
+    my $ref      = shift;
+    my $storage  = shift;
+    my $all      = {};
+    my $lockfile = "${storage}.lock";
+    my $hostkey  = find_hostkey($host);
 
+    die "no permission to read status file «${storage}»" unless -w ${storage};
```
rjuju, line 560: unless I'm missing something it's never going to work if the file isn't already existing?

Also, the error message is wrong.

I think it would be better to do the flock first, and then check for the file existence, create an empty one if needed (or make sure is_storable can handle missing file) or check for correct permissions if it exists.

check_pgactivity
```diff
@@ -500,20 +500,17 @@ sub list_services() {
...
-    open my $fh, '<', $storage;
-    flock($fh, LOCK_SH) or die "can't get shared lock on $storage: $!";
-
-    if ( defined *Storable::file_magic{CODE}
-         and Storable::file_magic( $storage )
-    ) {
-        close $fh; # release the shared lock
-        return 1;
+    if ( defined *Storable::file_magic{CODE} ) {
+        return 1 if defined Storable::file_magic( $storage );
```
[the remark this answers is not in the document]
ioguix: done

check_pgactivity
ioguix: You are right. I believe I fixed this.

I forced push after rebasing the branch, and added a TAP test for hit_ratio that is actually using the status file.

check_pgactivity
```diff
@@ -580,22 +584,27 @@ sub save($$$$) {
...
+    my $hostkey  = find_hostkey($host);
+    my $lockfile = "${storage}.lock";
     my $all;
 
     return undef unless -e $storage;
 
-    die "No permission to read status file «${storage}»" unless -r $storage;
+    die "can not read status file «${storage}»" unless -r $storage;
+
+    open my $fh, '<', $lockfile or die "can't open «${lockfile}»: $!";
```
rjuju, line 598: Shouldn't the open be `>` instead?  Otherwise it's unnecessary fragile against removal of the lock file.

check_pgactivity
```diff
@@ -552,24 +547,33 @@ sub find_hostkey($) {
...
+    my $lockfile = "${storage}.lock";
+    my $hostkey  = find_hostkey($host);
 
+    open my $fh, '>', $lockfile or die "can't open «${lockfile}»: $!";
+    flock($fh, LOCK_EX) or die "can't get exclusive lock on «${lockfile}»: $!";
 
-    die "File «${storage}» not recognized as a check_pgactivity status file.\n\n"
-        ."Please, check its path or move away this wrong file"
-        if -r $storage and not is_storable $storage;
+    if ( -e $storage ) {
```
rjuju, line 561: I think we should add an extra `!-d` check to be sure that the given storage is not a directory and raise some explicit error in that case?  I don't see any reason to refuse the use of symlinks there so those should be fine.

Same for the related code in `load()`.

check_pgactivity
```diff
@@ -500,20 +500,15 @@ sub list_services() {
 
 # Check wrapper around Storable::file_magic to fallback on
 # Storable::read_magic under perl 5.8 and below
+# WARNING: You must hold a lock on the lockfile **BEFORE** calling this sub
```
rjuju, line 503: and also mention that caller is responsible for checking that the given $storage exists, isn't a directory and is readable?

Nothing prevents outside modification of the file so an error could still happen, but we should at least be extra clear of what our own API is.

check_pgactivity
```diff
@@ -552,24 +547,33 @@ sub find_hostkey($) {
 #  * the ref of the structure to save
 #  * the path to the file storage
 sub save($$$$) {
```
rjuju, line 549: I think it would be good to document a bit more what `save()` is doing to make sure that you don't overwrite concurrent changes.

rjuju (changes requested):
I think there should also some changes in the readme nearby --status-file documentation about that new lock file.  And some warning about limitation with ntfs.

check_pgactivity
ioguix: I chose to open in read only because of the following note in the perl documentation of `flock`:

> Note that the fcntl(2) emulation of flock(3) requires that FILEHANDLE be open with read intent to use LOCK_SH and requires that it be open with write intent to use LOCK_EX.

Confirmed from the `man 2 fcntl`:

> In order to place a read lock, fd must be open for reading.  In order to place a write lock, fd must be open for writing.  To place both types of lock, open a file read-write.

So I chose to stay on the compatibility side of the fence, should the implementation of flock rely on fcntl on some user system.

check_pgactivity
ioguix: I added an early check, among all others before calling the service itself.

check_pgactivity
rjuju: Then shouldn't you adapt the rest of the calls accordingly?  For instance on line 574.

check_pgactivity
```diff
@@ -8994,6 +9019,13 @@ pod2usage(
 ) unless exists $services{ $args{'service'} };
 
 
+# Check that the given service exists.
```
rjuju, line 9022: Wrong comment.

rjuju (changes requested):

check_pgactivity
rjuju: Sorry I misread the  perldoc part, this is indeed correct :)

rjuju (approved):
LGTM!

PR #307, opened 2022-03-17
Fix threshold check and support interval for check_stat_snapshot_age

ioguix:
The threshold check was not strict enough, giving '1d' was ok where
the doc says this check only support raw values. In consequence,
check_pgactivity was failling with eg.:

  Argument '"2d" isn't numeric in numeric ge (>=)'

Worst, the check was returning "OK".

The fix supports interval as threshold value and fix the related tests.


rjuju (approved):
LGTM!

ioguix:
thx!

PR #308, opened 2022-04-06
add service "session_stats"

yhuelf:


check_pgactivity
```diff
@@ -3350,6 +3355,103 @@ sub check_btree_bloat {
...
+        my $sessions_fatal  = $new_sstats{$db}{'fatal'}  - $sstats{$db}{'fatal'};
+
+        # default to 1 sec if called twice in the same second
+        my $sec = ( $new_sstats{$db}{'ts'} - $sstats{$db}{'ts'} ) || 1;
+
+        my $session_rate = $sessions / $sec;
+        $instance_session_rate += $session_rate;
+
+        push @perfdata => (
+            [ "${db}_session_rate", sprintf( "%.2f ", $session_rate ), 'sessions/s' ],
```
ioguix, line 3436: I know there's another place where we format the perfdata precsion, but as these data are not aimed to humans, I wonder if we should keep this useless habit, or even fix it.

We might want to keep the best possible accuracy in perfdata and the formatting should happen on the frontend using them if really needed.

ioguix (changes requested):

check_pgactivity
```diff
@@ -3350,6 +3355,103 @@ sub check_btree_bloat {
...
+        my $sessions = $new_sstats{$db}{'sessions'} - $sstats{$db}{'sessions'};
+        my $session_time = $new_sstats{$db}{'session_time'} - $sstats{$db}{'session_time'};
+        my $active_time  = $new_sstats{$db}{'active_time'}  - $sstats{$db}{'active_time'};
+        my $idle_in_tx_time  = $new_sstats{$db}{'idle_in_tx_time'}  - $sstats{$db}{'idle_in_tx_time'};
+
+        my $sessions_killd  = $new_sstats{$db}{'killd'}  - $sstats{$db}{'killd'};
+        my $sessions_aband  = $new_sstats{$db}{'aband'}  - $sstats{$db}{'aband'};
+        my $sessions_fatal  = $new_sstats{$db}{'fatal'}  - $sstats{$db}{'fatal'};
+
+        # default to 1 sec if called twice in the same second
```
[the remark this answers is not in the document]
yhuelf: I agree with you :-)

check_pgactivity
```diff
@@ -3350,6 +3355,103 @@ sub check_btree_bloat {
...
+        push @perfdata => (
+            [ "${db}_session_rate", sprintf( "%.2f ", $session_rate ), 'sessions/s' ],
+
+            [ "${db}_session_time",  sprintf( "%i", $session_time ), 'ms' ],
+            [ "${db}_active_time",   sprintf( "%i", $active_time ),  'ms' ],
+            [ "${db}_idle_in_transaction_time", sprintf( "%i", $idle_in_tx_time ), 'ms' ],
+
+            [ "${db}_sessions_killed",    sprintf( "%i", $sessions_killd ) ],
+            [ "${db}_sessions_abandoned", sprintf( "%i", $sessions_aband ) ],
+            [ "${db}_sessions_fatal",     sprintf( "%i", $sessions_fatal ) ]
```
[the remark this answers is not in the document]
yhuelf: You're right. Fixed

check_pgactivity
yhuelf: Agreed too :-) Fixed

check_pgactivity
```diff
@@ -85,6 +85,7 @@ my $PG_VERSION_100 = 100000;
 my $PG_VERSION_110 = 110000;
 my $PG_VERSION_120 = 120000;
 my $PG_VERSION_130 = 130000;
+my $PG_VERSION_140 = 140000;
```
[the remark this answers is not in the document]
yhuelf: Yes, it would have been cleaner, I'm going to rebase my branch then

ioguix (approved):
This patch is sounds to me.

@rjuju, any comments/objections?

check_pgactivity
```diff
@@ -3351,6 +3355,100 @@ sub check_btree_bloat {
...
+    %sstats = %{ load( $hosts[0], 'session_stats', $args{'status-file'} ) || {} };
+
+    @rs = @{ query( $hosts[0], $sql ) };
+
+    $new_sstats{$_->[1]} = {
+        'ts'               => $_->[0],
+        'sessions'         => $_->[2],
+        'session_time'     => $_->[3],
+        'active_time'      => $_->[4],
+        'idle_in_tx_time'  => $_->[5],
```
jrouhaud, line 3408: nitpicking: everywhere else in the code we use xact for transaction, not tx, so it would be better to be consistent here.

check_pgactivity
```diff
@@ -3350,6 +3355,100 @@ sub check_btree_bloat {
...
```
rjuju, line 3408: nitpicking: everywhere else in the code we use xact for transaction, not tx, so it would be better to be consistent here.

yhuelf: fixed

ioguix:
Thanks!

Issue #306, opened 2022-03-15
Change oldest_idlexact to use state_change to detect if transaction is **really** idle


frost242:
Hello,

My monitoring alerts me about idle transaction running for more than 1h. However, when looking at the detail, the transaction is open since some time but the transaction remains active, as shown here :
```
datid            | 455479
pid              | 9769
usesysid         | 17788
client_port      | 36804
backend_start    | 2022-03-15 13:00:20.119499+01
xact_start       | 2022-03-15 14:57:55.658817+01
query_start      | 2022-03-15 16:14:41.999261+01
state_change     | 2022-03-15 16:14:42.192581+01
wait_event_type  | Client
wait_event       | ClientRead
state            | idle in transaction
backend_xid      | 2018359
backend_xmin     | 2018359
```

What do you think about using the state_change column to check if a transaction is really idle ?

Krysztophe:
It depends.

In your case, this is not a dead session, it is not stuck.

But it still is a problem if it prevents the autovacuum to do its job. You don't want a buggy script that makes an update every hour, and never commits.

Do we need 2 different thresholds? I don't like it.

BTW: longest_query uses  state_change already

frost242:
Regarding the vacuum issue, the problem is triggered by any transaction open for too long, idle or not.

IMHO, after some time off regarding the development of check_pga, when I read "oldest_idlexact", I feel that the probe should trigger an alert if a transaction is really idle. But maybe check_pga lacks a service that would be called oldest_xact that behaves like the current oldest_idlexact and triggers an alert if a transaction remain active for too long - longest_query has not the same goal.

rjuju:
I agree with @frost242.  Anyone using that service will think that it's checking for transaction that have been idle for more than the threshold, not transaction that have been started before the threshold and happens to be idle, even for only a millisecond, when we check.  Unfortunately our documentation is entirely failing to describe what are the thresholds actually doing.

Note that if you have problem with bloat / xmin retention and want to troubleshoot that, transactions are just one possible source of problem, and you should have better success using the oldest_xmin service I added a bit more than a year ago.

frost242:
Hello,
I will provide a patch this week.
Kind regards

PR #311, opened 2022-05-03
Fix check_oldest_idlexact to use state_change

frost242:
Fix check_oldest_idlexact to use state_change instead of xact_start to
calculate the idle time. The previous behavior leads to warn for active
long transaction.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -5722,7 +5722,7 @@ Above 9.2, it supports C<--exclude> to filter out connections. Eg., to
 filter out pg_dump and pg_dumpall, set this to 'pg_dump,pg_dumpall'.
 
 Before 9.2, this services checks for idle transaction with their start time.
-Thus, the service can mistakenly take account of transaction transciently in
+Thus, the service can mistakenly take account of transaction transiently in
 idle state.
 From 9.2 and up, the service checks for transaction that really had no activity
 since the given thresholds.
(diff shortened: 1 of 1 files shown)


frost242:
This fixes issue  #306

frost242:
Please wait until tomorrow afternoon, I want to make an extensive test of this at work.

frost242:
In a next PR

ioguix (changes requested):
It deserve some user facing documentation explaining the diff between the what alert triggers on before/after 9.2, where `state_change` appears.

Plus, could you add `fix #306` in the commit message?

Thanks

frost242:
Some documentation added, but I'm not very happy with the result. Do you have a better wording ?
If OK, I'll squash commit and add reference for this issue.

check_pgactivity
```diff
@@ -5721,6 +5721,12 @@ This service supports both C<--dbexclude> and C<--dbinclude> parameters.
 Above 9.2, it supports C<--exclude> to filter out connections. Eg., to
 filter out pg_dump and pg_dumpall, set this to 'pg_dump,pg_dumpall'.
 
+Before 9.2, this services checks for idle transaction with their start time.
+Thus, the service can wrongly can take account of transaction transciently in
```
ioguix, line 5725:
```suggestion
Thus, the service can wrongly take account of transaction transiently in
```

check_pgactivity
```diff
@@ -5721,6 +5721,12 @@ This service supports both C<--dbexclude> and C<--dbinclude> parameters.
...
+idle state.
+From 9.2 and up, the service checks for transaction that had no activity since
```
ioguix, line 5727:
```suggestion
From 9.2 and up, the service checks for transaction that really had no activity since
```

frost242:
A bit better I think. Thought ?

frost242:
Is that better ?

ioguix (approved):

PR #313, opened 2022-05-19
Fix small typo into thresholds_re var name

l00ptr:
We simply replace and fix the typo threshods_re -> thresholds_re.


ioguix:
Sounds good to me. Thank you!

Issue #323, opened 2022-07-11
Suggest a periodicity for all services in the description of each service


Krysztophe:
That would be helfpul for beginners.

Examples:
```
check_uptime:
.... 
Suggested periodicity: 5 min

max_freeze_age:
...
Suggested periodicity: 1 hour
```

I may try to write the PR (and I'm aware that that this can lead to very long discussions).

rjuju:
I don't think this would be something sensible to do.  I'm afraid it will lead to people not really trying to think about what they actually want to achieve but just follow some abstract guidelines.  Also, wouldn't the suggestion usually depends on whether your primary goal is alerting vs graphing?

I think that we should simply warn about the  services that can be costly (like the bloat), so people have useful information.

rjuju:
How would you even define "failing"?

I see on your PR that you just claim that some periodicity are good, but without any reasoning why.  Even if that's true, I don't think it's helpful as is without some clarification why this value is suggested.

Note: saying that using backup-specific services doesn't seem like periodicity related.

Krysztophe:
https://github.com/OPMDG/check_pgactivity/pull/324#issuecomment-1831718111

PR #324, opened 2022-07-11
Doc: suggest periodicities

Krysztophe:
Documention: suggest a default periodicity for each service for users who have clue.


Krysztophe:
Linked to #323

Krysztophe:
@ioguix  not convinced, I give up

Issue #326, opened 2022-08-18
bug in 2.6 on rhel7


joernott:
Hello,
we are using check_pgactivity on RHEL7.9 (using the RPMs provided by the postgres project). When we updated 2.5 (which works fine) to 2.6, we encountered the following error when creating status files:
```
can't open «/var/cache/icinga2/icinga_bgwriter.status.lock»: No such file or directory at /usr/lib64/nagios/plugins/check_pgactivity line 619.
```
The parent folder /var/cache/icinga2 is owned by the icinga2 user executing the check and has rwx rights for that user and using "sudo -u $MONITORING_USER touch /var/cache/icinga2/icinga_bgwriter.status.lock" works fine.

This is similar for all status file, this is just one example for the "icinga" database and the bgwriter check. 

I solved our current problem by downgrading back to check_pgactivity 2.5 but I'd rather use the latest released version if possible.

rjuju:
Hi,

Indeed the version 2.6 is bugged :(  The new locking approach triggers that problem when upgrading from an earlier version, as `load()` will see a storage file, but won't be able to get a shared lock on the lock file since it won't exist.

BTW, is there any reason to use a status files rather one for everything?  Creating an empty lock file should fix the problem, but clearly if you need dozen (or more) of them that's not a pleasant thing to do.

@ioguix I remember I mentioned that problem a couple of times in some versions of the patch (https://github.com/OPMDG/check_pgactivity/pull/303#discussion_r804345650 and https://github.com/OPMDG/check_pgactivity/pull/303#discussion_r814663680), but forgot to recheck this in the final versions.  You can reproduce this easily by removing the lock file, I guess this should be added to the regression tests.  We just need to double check the lock file existence in `load()` and create an empty one if needed.

joernott:
I am using a status file per check, as I didn't bother to analyze, what every check does with that file. Most of them store check specific info (how far have I read the log file, what was the last value of a counter etc.). So, in general, having different checks using the same file causes a lot of trouble.

For the moment, I am using 2.5 again, but as we manage our environments via puppet, creating lots of files on lots of servers isn't that much of an effort.

ioguix:
Hi guys,

@joernott: thanks for the report!
@rjuju: sorry, your comment has been lost in other discussions :/

What about the following fix?

~~~diff
diff --git a/check_pgactivity b/check_pgactivity
index 2aeb33a..192a53c 100755
--- a/check_pgactivity
[...]

PR #328, opened 2022-09-08
Add test file dedicated to regression tests.

ioguix:
The current only regression test is about #326


t/02-regression_tests.t
```diff
@@ -0,0 +1,52 @@
...
+
+# == Regression test for #326 ==
+# check_pga should not complain when using an existing status file missing
+# without an existing lock file.
+my $pga_data = "$TestLib::tmp_check/tmp-status-file.data";
+
+# Create an empty status file
+store( {}, $pga_data);
+# make there's no lock file leftovers from previous tests...
+unlink "${pga_data}.lock";
```
rjuju, line 38: I was thinking it would be better to test the other way:
- test that no status file exists
- run a test that relies on the status file
- test that both the expectedstatus file and lock file exist
- remove the lock file
- run the test again, which should work
- possibly test again that the lock file exists

rjuju (changes requested):
Nitpicking, but "regression_tests" seems a bit abstract.  Maybe "general", or "infrastructure" or something like that for the name of the test?

t/02-regression_tests.t
ioguix: I first wrote the tests with two different call to check_pga, to create the status file, remove the lock, and re-run the test again. 

But I finally choose the fast lane as I was wondering what this was buying us in the end of the day? Just a test to make sure check_pga is creating its status file when missing (your second step)?

ioguix:
renamed as 't/02-general.t'

t/02-regression_tests.t
rjuju: It's just to be thorough, and validate that the 2 possible scenario (totally empty environment and environment upgraded from a previous version) work as intended.  Even if it's really unlikely that it ever gets broken it doesn't cost much to do so.

t/02-general.t
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/perl
+# This program is open source, licensed under the PostgreSQL License.
+# For license terms, see the LICENSE file.
+#
+# Copyright (C) 2012-2022: Open PostgreSQL Monitoring Development Group
+
+# This file gather various regression tests against the same cluster to avoid
```
rjuju, line 7: nit: gather**s**

t/02-general.t
```diff
@@ -0,0 +1,52 @@
...
+my $node = pgNode->get_new_node('prod');
+
+$node->init;
+
+$node->append_conf('postgresql.conf', 'stats_block_level = on')
+    if $node->version < 8.3;
+
+$node->start;
+
+### Begin of tests ###
```
rjuju, line 28: nit: Begin**ning**

t/02-general.t
```diff
@@ -0,0 +1,52 @@
...
+
+# == Regression test for #326 ==
+# check_pga should not complain when using an existing status file missing
```
rjuju, line 31: the "missing" should be removed?

t/02-general.t
```diff
@@ -0,0 +1,52 @@
...
+# without an existing lock file.
+my $pga_data = "$TestLib::tmp_check/tmp-status-file.data";
+
+# Create an empty status file
+store( {}, $pga_data);
+# make there's no lock file leftovers from previous tests...
```
rjuju, line 37: nit: make **sure that** ...?

t/02-general.t
```diff
@@ -0,0 +1,52 @@
...
+unlink "${pga_data}.lock";
+
+# Trigger the failure
```
rjuju, line 40: should be rephrased.  Maybe "Trigger the call that failed before fixing bug 326" or something like that?

t/02-general.t
```diff
@@ -0,0 +1,52 @@
...
+$node->command_checks_all( [
+    './check_pgactivity', '--service'     => 'hit_ratio',
+                          '--username'    => getlogin,
+                          '--format'      => 'human',
+                          '--status-file' => $pga_data,
+                          '--warning'     => '101%',
+                          '--critical'    => '0%'
+    ],
+    0, [], [ qr/^$/ ], 'No error should occurs'
```
rjuju, line 49: nit: should occur

rjuju (changes requested):
Nit for comments

t/02-general.t
ioguix: done

t/02-general.t
ioguix: reworded as «Trigger the failure described in issue #326»

t/02-regression_tests.t
ioguix: done

rjuju (approved):
LGTM!

ioguix:
thanks @rjuju !

PR #330, opened 2022-09-09
check_last_maintenance : filter out small tables

yhuelf:
The idea is to avoid false positives generated by empty or very small tables.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -458,6 +458,8 @@ my %args = (
     'wal_buffers'           => undef,
     'checkpoint_segments'   => undef,
     'effective_cache_size'  => undef,
+    'vacuum_table_min_size' => 10 * 1024 * 1024,
+    'analyze_table_min_size' => 1024 * 1024,
     'no_check_autovacuum'   => 0,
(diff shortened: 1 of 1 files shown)


ioguix:
The last_vacuum & last_analyze services are supposed to check if the planned maintenances succeed.

If you don't have planned maintenances, you don't need to check them, you don't need to deploy this service, no matter how large are your tables.

This should probably be documented, so @Krysztophe just did it in PR #366.

However, when studying this issue, we found partitioned tables were raising errors even after being vacuumed. The issue #365 has been created to track this.

ioguix:
@Krysztophe insist to say:

> on empty tables, it doesn't matter to run vacuum because it's almost cost free. So no need to optimize this use case.

yhuelf:
Actually, I disagree with you. You can have planned maintenances, but some people have working tables which are created / destroyed regularly. We could avoid the false positives with an `--exclude` parameter (#369), but filtering out very small tables (with less than 50 tuples) would be beneficial.

(that being said, the default thresholds of this PR are too high)

PR #331, opened 2022-09-14
Add a new service pgss_dealloc

anayrat:
Hello!
This [conversation on hackers](https://www.postgresql.org/message-id/20220912131800.ttcgyorpc3wvzwsq%40jrouhaud)  remind me that pg_stat_statements deallocs should me monitored.

I suggest adding such service.

Cheer


anayrat:
If the number of dealloc per second is too low, we can change it to number of dealloc per millisecond.

check_pgactivity
```diff
@@ -6547,6 +6550,114 @@ sub check_pg_dump_backup {
...
+    };
+
+    @hosts = @{ parse_hosts %args };
+
+    pod2usage(
+    -message => 'FATAL: you must give only one host with service "commit_ratio".',
+    -exitval => 127
+    ) if @hosts != 1;
+
+    is_compat $hosts[0], 'check_pg_stat_statements_dealloc', $PG_VERSION_140 or exit 1;
```
rjuju, line 6607: Note that having postgres 14 doesn't mean that you updated the pg_stat_statements extension to get the needed field.

rjuju (commented):
I'm a bit dubious about this service.  I agree that this is something you should keep you eyes on, but I don't think it would play very well as a check here.

The main problem is that if you schedule the check too frequently it wil be a bit useless.  For instance, if you schedule it every 5 minutes, how do you differentiate from "there was once 1 deallocate and then none" from "there is 1 deallocate every 5 minutes" from this service point of view?  The only way to know if there's really a problem is either:

1) the service is constantly raising a problem
2) the frequency of the service moving from ok to problem is high

But if you're in the first case it's likely that the global performance will immediately drop down by a huge factor, so it's unlikely that you won't notice there's a problem.  And the second isn't a good way to spot a problem.

The fact that you only return (and handle thresholds as) a rate and not also the raw number probably exacerbates this problem.

check_pgactivity
anayrat: Indeed, I will add a test to check pgss' version.

anayrat:
Yeah, I shouldn't report rate as perfdata. I will change it to a counter. But for the threshold, I don't see other way.
We can't use a threshold as a raw value. For example, if you have 1 dealloc every minute, once you reach 100 (if it is your threshold). The check be critical even if there is no more deallocation.

My idea is to, first graph the dealloc rate. For example, I will give you a mean rate of 100 dealloc between 5 minutes. Then, you add a threshold at 500. That means if you reach this threshold, your workload has changed. And, you should understand why you have an increase in dealloc rate **before** hitting production issue.

ioguix:
Hi guys,

Instead of working on rates, the threshold can apply on delta since last call. I think this is what @anayrat explains in his last message, is it?

So you plan to keep working on it @anayrat?

anayrat:
Hello,
Yes, I can use the delta since last call as a threshold.
I will work on this.

check_pgactivity
anayrat: I added several tests:
* pg_stat_statements version must be above or equal 1.9
* pg_stat_statements has been created on target database
* pg_stat_statements has been loaded in shared_preload_libraries

anayrat:
Hello,
I added more checks, replaced rate by dealloc delta and rephrased service description.
If you want to test : 
Run a pgbench with
```
\set id random(1,1000000)
set client.id = :id
```
And enable `pg_stat_statements.track_utility` with a low `pg_stat_statements.max`

check_pgactivity
```diff
@@ -6700,6 +6703,133 @@ sub check_pg_dump_backup {
...
+    -message => 'FATAL: you must give only one host with service "pgss_dealloc".',
+    -exitval => 127
+    ) if @hosts != 1;
+
+    is_compat $hosts[0], 'check_pg_stat_statements_dealloc', $PG_VERSION_140 or exit 1;
+
+    @rs = @{ query( $hosts[0], $pgss_version_query ) };
+
+    # No row:
+    return status_unknown( $me, [ 'Ensure pg_stat_statements extension has been created on target Database.' ] ) if (not defined $rs[0]);
```
yhuelf, line 6773:
```suggestion
    return status_unknown( $me, [ 'Ensure that the pg_stat_statements extension has been created on the target database.' ] ) if (not defined $rs[0]);
```

check_pgactivity
```diff
@@ -6700,6 +6703,133 @@ sub check_pg_dump_backup {
...
+
+    $pgss_version = $rs[0][0];
+    $pgss_version =~  /(\d+).(\d+)/;
+    return status_unknown( $me, [ 'pg_stat_statements version must be above or equal 1.9' ] ) if (not ($1 >=1 and $2 >= 9));
+
+    return status_unknown( $me, [ 'Ensure pg_stat_statements has been loaded in shared_preload_libraries.' ] ) if ($rs[0][1] ne 't');
```
yhuelf, line 6779:
```suggestion
    return status_unknown( $me, [ 'Ensure that the pg_stat_statements has been loaded in shared_preload_libraries.' ] ) if ($rs[0][1] ne 't');
```

yhuelf (commented):
typos

yhuelf:
I have tested, it works fine.

anayrat:
Thanks @yhuelf, I applied your suggestions.

gleu:
> Thanks @gleu, I added your patch and rebased the PR.

You're welcome. And I pushed your patch on master. PR done :)

PR #332, opened 2022-09-28
Fix backends_status to use state_change for pg>=9.2

frost242:
Previously, backends_status used xact_start to check if a transaction is waiting for a lock or is idle for a given amount of time. This leads to false positive alerts when a transaction becomes transciently idle.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -2516,7 +2516,7 @@ sub check_backends_status {
                 ELSE s.state
               END,
               extract('epoch' FROM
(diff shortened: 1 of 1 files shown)


frost242:
Can we merge this fix ? It works nicely in production at work. I'll try to provide another patch for the regression tests later.

rjuju (approved):
LGTM, and that's already what we're doing for check_oldest_idlexact for pg 9.2+ so +1 to report a better metric.

PR #341, opened 2022-12-06
Fix commit cb0e5f4

frost242:
A typo was done in commit cb0e5f4.
Also, add documentation about required privileges.


rjuju (approved):
LGTM!

Issue #345, opened 2022-12-07
check_backends_status: accounting of active session with wait event


frost242:
Hello,

Since the introduction of wait events report in pg_stat_activity (pg9.6), we account active sessions having a wait_event_type other than 'Lock' in category "other wait event".
For example :
```
postgres=# select state, wait_event_type, query from pg_stat_activity WHERE state = 'active';
-[ RECORD 1 ]---+-----------------------------------------------------------------------------------
state           | active
wait_event_type | Timeout
query           | SELECT pg_sleep(60);
-[ RECORD 2 ]---+-----------------------------------------------------------------------------------
state           | active
wait_event_type | 
query           | select state, wait_event_type, query from pg_stat_activity WHERE state = 'active';
```

When running check_backends_status, the running `SELECT pg_sleep(60)` is not accounted as active :
```
postgres@nas:~$ /usr/lib/nagios/plugins/check_pgactivity -s backends_status -w active=1 -c active=4 -F human
Service        : POSTGRES_BACKENDS_STATUS
Returns        : 1 (WARNING)
Message        : 1 active
Perfdata       : active=1 warn=1 crit=4 min=0 max=100
Perfdata       : idle=1
Perfdata       : other wait event=1
Perfdata       : oldest other wait event=2.052000s
```

The idle=1 perfdata is dubious, will dive into it.

The query used returns the following result :
```
      state       |  extract  | current_setting |                                    query                                     
------------------+-----------+-----------------+------------------------------------------------------------------------------
 other wait event | 10.873000 | 100             | SELECT pg_sleep(60);
 active           |  0.000000 | 100             | SELECT CASE                                                                 +
                  |           |                 |                 WHEN s.wait_event_type = 'Lock' THEN 'waiting for lock'     +
                  |           |                 |                 WHEN s.query = '<insufficient privilege>'                   +
                  |           |                 |                     THEN 'insufficient privilege'                           +
                  |           |                 |                 WHEN s.state IS NULL THEN 'undefined'                       +
                  |           |                 |                 WHEN s.wait_event_type IS NOT NULL                          +
                  |           |                 |                        AND s.wait_event_type NOT IN ('Client', 'Activity')  +
                  |           |                 |                      THEN 'other wait event'                                +
                  |           |                 |                 ELSE s.state                                                +
                  |           |                 |               END,                                                          +
                  |           |                 |               extract('epoch' FROM                                          +
[...]

PR #346, opened 2022-12-07
Fix check_backends_status to simplify session accounting

frost242:
Active backends with wait_event_type other then Active ou Client were not accounted as active.
Add a bit more debug output and a regression test for it.

Fix issue #345

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -2490,7 +2490,6 @@ sub check_backends_status {
         'disabled'                      => [0, 0],
         'undefined'                     => [0, 0],
         'insufficient privilege'        => [0, 0],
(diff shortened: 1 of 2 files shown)


rjuju:
Did you really want to keep the 2 first commits?  I don't think it's really useful to have those apart when actually working on a patch, and you can always temporarily add such while doing so.

In general, +1 for removing the "other wait event" state.

frost242:
> Did you really want to keep the 2 first commits? I don't think it's really useful to have those apart when actually working on a patch, and you can always temporarily add such while doing so.

Done.

> In general, +1 for removing the "other wait event" state.

Thanks. PR is ready.

Issue #355, opened 2023-09-11
pg_dump_backup : error is --path is a directory


Krysztophe:
I've lost too much time searching why the service does not read any file.

  * Option 1: error if `--path` is a directory 
  * Option 2: if a directory, simply add `*`.

I would prefer option 1.
 
As a second thought, `--debug` should trace a bit more which files it is reading. Debugging regex is painful enough to avoid doing it in the dark.

PR #356, opened 2023-09-11
#355: pg_dump_backup : error if --path is a directory

Krysztophe:
Add an error message if `--path` is a directory and not a path with wildcards.
(a mistake so easy and obvious that documenting is not enough).

Added 2 debug statements that were really missing for people trying to understand which files were found.


frost242:
LGTM

PR #362, opened 2023-11-29
(#325, last_maintnenance) Raise alerts on threshold change

### LINKED ISSUE #325: last vacuum/analyze updated thresholds not enforced
The last vacuum/analyze check track table activity using a `hash(insert||update||delete)` to detect write activity between two run and avoid useless alerts.

When changing the alert threshold below the current table last maintenance, if the table has no activity, the check will not raise an alert, but the perfdata will still report last vacuum greater than the threshold. This should probably be documented as this might be surprising.

Here is an exemple:

~~~
$ perl -l ./bin/check_pgactivity -F human \
    --status-file /tmp/check_pga.data     \
    --service last_vacuum                 \
    --dbservice main                      \
    --warning '28d'                       \
    --critical '35d'
Service        : POSTGRES_LAST_VACUUM
Returns        : 0 (OK)
Message        : 3 database(s) checked
...
Perfdata       : db3=2088224.763214s warn=2419200 crit=3024000
Perfdata       : db3 vacuum=0
Perfdata       : db3 autovacuum=0


$ perl -l ./bin/check_pgactivity -F human \
    --status-file /tmp/check_pga.data     \
    --service last_vacuum                 \
    --dbservice main                      \
    --warning '7d'                        \
    --critical '35d'
Service        : POSTGRES_LAST_VACUUM
Returns        : 0 (OK)
Message        : 3 database(s) checked
...
Perfdata       : db3=2088234.258004s warn=604800 crit=3024000 <======= HERE
Perfdata       : db3 vacuum=0
Perfdata       : db3 autovacuum=0
~~~

Krysztophe:
When there is no write, the last_analyze and last_vacuum raised no alert when crossing the threshold. This is okay. But lowering the threshold did not lead to an alert.

Add the thresholds in the MD5 stored in the status file to force a new test when they change. That may lead to a new alert on totally inactive server when check_pgactivity is upgraded.

Close #325

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -4976,14 +4976,29 @@ sub check_last_maintenance {
     my @dbinclude  = @{ $args{'dbinclude'} };
     my @dbexclude  = @{ $args{'dbexclude'} };
     my $me         = 'POSTGRES_LAST_' . uc($type);
+
+    # warning and critical are mandatory.
+    pod2usage(
+        -message => "FATAL: you must specify critical and warning thresholds.",
+        -exitval => 127
+    ) unless defined $args{'warning'} and defined $args{'critical'} ;
+
+    pod2usage(
+        -message => "FATAL: critical and warning thresholds only acccepts interval.",
+        -exitval => 127
+    ) unless ( is_time( $args{'warning'} ) and is_time( $args{'critical'} ) );
+
+    $c_limit = get_time $args{'critical'};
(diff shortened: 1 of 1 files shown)


ioguix (changes requested):
Just a small formatting suggestion, then you can merge!

Krysztophe:
@ioguix  said I could merge

Issue #365, opened 2023-11-30
last_vacuum : ignore partitions


Krysztophe:
Partitionned tables are empty. You can vacuum them but `pg_stat_user_tables.last_(auto)vacuum` is always empty. So let's exclude `relkind='p'` from `last_vacuum`.

This is different for last_analyze, partitionned tables can have `last_analyze` set.

PR #368, opened 2024-05-15
Ignore partition table in check_last_maintenance

anayrat:
Partitioned tables are empty, thus last_* are null.

Fix https://github.com/OPMDG/check_pgactivity/issues/365

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -5053,9 +5053,11 @@ sub check_last_maintenance {
                   FROM pg_catalog.pg_stat_database
                   WHERE datname = current_database()
                 )
-            FROM pg_stat_user_tables
+            FROM pg_stat_user_tables a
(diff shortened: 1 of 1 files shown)


Krysztophe:
We should not ignore `relkind = 'p'`  for last_analyze, right?

robinportigliatti:
> We should not ignore `relkind = 'p'` for last_analyze, right?

No we should not

SkyMarshall did another PR for this problem : https://github.com/OPMDG/check_pgactivity/pull/379

yhuelf:
@blogh's suggestion looks good to me.

gleu:
@anayrat Can you fix your PR with @blogh suggestion? Thanks.

frost242:
Thanks @anayrat

Issue #371, opened 2024-09-04
longest_query bails out because of a walsender session


frost242:
Hello,

We use debezium to track changes in some DBs and longest_query throw an error because of the walsender session is marked as inactive since for a long time :
```
datid            | 17818
datname          | ....
pid              | 2320507
leader_pid       | [null]
usesysid         | 21762
usename          | ....
application_name | PostgreSQL JDBC Driver
client_addr      | ....
client_hostname  | [null]
client_port      | ....
backend_start    | 2024-08-09 18:07:28.693025+02
xact_start       | [null]
query_start      | 2024-08-09 18:07:28.751659+02
state_change     | 2024-08-09 18:07:28.751697+02
wait_event_type  | Client
wait_event       | WalSenderWaitForWAL
state            | active
backend_xid      | [null]
backend_xmin     | [null]
query_id         | [null]
query            | START_REPLICATION SLOT "..." LOGICAL F/540E9AB8 ("proto_version" '1', "publication_names" '...', "messages" 'true')
backend_type     | walsender
```

We should ignore walsender backends in this query.

ioguix:
> because of the walsender session is marked as inactive since for a long time

You meant, «is marked as **active** since a long time», right ?

I agree walsenders should be ignored.

frost242:
yes, I mean that.

frost242:
Commited

PR #372, opened 2024-09-11
longest_query only account client backends, bgworkers and autovacuum workers

frost242:
Some walsender sessions emit a "START REPLICATION" query and are accounted as active queries, while there's no impact on the system and the replication is up and running.

Fixes issue #371

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -2323,6 +2323,9 @@ Critical and Warning thresholds accept either a raw number or a percentage (eg.
 between the cluster parameters C<max_connections> and
 C<superuser_reserved_connections>.
 
+Autovacuum workers and walsender process are ignored. To check autovacuum, use
+the autovacuum service.
+
 Required privileges: an unprivileged user only sees its own queries;
 a pg_monitor (10+) or superuser (<10) role is required to see all queries.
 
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -5489,7 +5489,24 @@ sub check_longest_query {
...
+            FROM pg_database AS d
+            LEFT JOIN (
+                SELECT datname, query,
+                    extract('epoch' FROM
+                        date_trunc('second', current_timestamp-state_change)
+                    ) AS elapsed,
+                    application_name
+                FROM pg_stat_activity
+                WHERE state = 'active'
+		  AND backend_type IN ('client backend', 'background worker', 'autovacuum worker')
```
ioguix, line 5506: I wonder if we should remove autovacuum worker here.

There's already a service relating to autovacuum, this check is about is about long running user query and autovacuum is designed to be slowly-ish, in the background anyway.

ioguix (changes requested):

check_pgactivity
frost242: I asked myself this question, but didn't think about the redundancy with the autovacuum specific check. Let's remove them and add a comment suggesting to use the autovacuum service.

PR #378, opened 2025-04-08
Fix annoying race condition in archiver service

ioguix:
Function pg_current_wal_lsn() can return a LSN right at the begining of a new WAL segment, like XX/YY000000. Also, archiver service query always tries to check for how long the next segment exists using pg_stat_file(). No matter if it is waiting for archiver or the actual current WAL.

However, pointing to XX/YY000000 doesn't mean the WAL already exists if no write occurs to actualy create the WAL. This leads to annoying UNKNOWN status with the following error:

  Info:    CHECK_PGACTIVITY UNKNOWN: Query failed !
  ERROR:  could not stat file "pg_wal/000000010000076500000095":
          No such file or directory

Function pg_stat_file() can ignore missing file since 9.5, so just ignore and report 0s, which is actually what we already do anyway when no WAL are waiting to archive.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -2026,7 +2026,7 @@ sub check_archiver {
                         (pg_stat_file('pg_wal/'||pg_walfile_name(
                            (to_hex((last_archived_off+1)/4294967296)
                            ||'/'||to_hex((last_archived_off+1)%4294967296))::pg_lsn
-                      ))).modification ))
+                      ), true)).modification ))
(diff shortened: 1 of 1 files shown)


blogh:
lgtm

blogh:
Note: there as been changes to `pg_walfile_name()` when the LSN was on a segment boundary
I don't think it changes anything to this patch though.

https://github.com/postgres/postgres/commit/344afc77690b4cef2605420d67c1d1ceb28760cc

PR #393, opened 2025-10-17
Check if server is primary for last_maintenance

### LINKED ISSUE #359: Last Vacuum Always Errors on Hot Standby Cluster
When using the last_vacuum check on a hot standby cluster a critical alert is always raised. This appears to be caused by the fact that a hot standby does not maintain metadata for when a vacuum job is replicated from the primary to the standby. When running the same check on the primary cluster there is no alert generated.

In attempting to fix this internally I have modified `sub check_last_vacuum` to borrow the code from `sub check_is_hot_standby` and only running the check if the cluster isn't a hot standby cluster. What would be the best way to address this issue, and would it be possible to get a more appropriate fix added directly to check_pgactivity?

Below is what I did to `sub check_last_vacuum` which appears to work for my needs, but I am not a programmer so forgive any stupidity that you find.

```
sub check_last_vacuum {
    my @rs;
    my @hosts;
    my %args          = %{ $_[0] };
    my $me            = 'POSTGRES_CHECK_LAST_VACUUM';
    my %queries       = (
        $PG_VERSION_90 => q{SELECT pg_is_in_recovery()}
    );

    @hosts = @{ parse_hosts %args };

    pod2usage(
        -message => 'FATAL: you must give only one host with service "check_last_vacuum".',
        -exitval => 127
    ) if @hosts != 1;

    is_compat $hosts[0], 'is_hot_standby', $PG_VERSION_90 or exit 1;
    @rs = @{ query_ver( $hosts[0], %queries )->[0] };

    return check_last_maintenance( 'vacuum', @_ ) if $rs[0] eq "f";
    return status_ok( $me, [ "Cluster is hot standby, skipping check" ] );
}
```

gleu:
Fix #359.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -5103,20 +5103,32 @@ sub check_last_maintenance {
         },
         # Starting with 9.1, we can add the analyze/vacuum counts
         $PG_VERSION_91 => qq{
-            SELECT coalesce(max(
+            SELECT
+                pg_is_in_recovery()::int AS is_in_recovery,
+                CASE WHEN NOT pg_is_in_recovery() THEN
+                  coalesce(max(
                        coalesce(extract(epoch FROM
                          current_timestamp -
                                  greatest(last_${type}, last_auto${type})),
                         'infinity'::float)),
-                   'NaN'::float),
-                coalesce(sum(${type}_count), 0) AS ${type}_count,
-                coalesce(sum(auto${type}_count), 0) AS auto${type}_count,
-                (
-                  SELECT md5(tup_inserted::text||tup_updated::text||tup_deleted::text||'$c_limit $w_limit'
+                   'NaN'::float)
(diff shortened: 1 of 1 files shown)


gleu:
Well, I used the same trick as the is_replay_paused service. I checked all the source code. It uses my way, not yours. Better stay on the same trick, right?

gleu:
Oh, no, got it. It subtracts previous values from current ones. So this is good to go. At least for me :)

gleu:
I see no complaints. Barring any objections, I'll merge this PR on sunday (and close the other one).

gleu:
You're right. I've seen some tests failing, obviously the services I've changed. I'll get a look at it tomorrow.

gleu:
It took me some time but I finally found the bug. So, should be fixed :)

PR #397, opened 2025-10-20
Only run actions on push and truncate log output

anayrat:
We have an issue with temp_files tests which generate a log of 10 million line. Let's truncate it.
Also, avoid running tests twice. Push is enough.


frost242:
It's quite interesting to have this feature available on pull requests. We better should see how we can prevent generating so much logs.

gleu:
Is it still a draft? Would be nice to apply this.

anayrat:
Sorry, I think I fixed the conflict

gleu:
Sounds good to me.

gleu:
Thank you.

PR #406, opened 2026-01-16
Allow warnings from the settings service to expire

### LINKED ISSUE #405: Settings service: timeout to avoid `--save`
After the settings service triggers an alert, it must be reset with `--save`.

It is a tedious manual operation, and more complicated with named accounts different from the monitoring account (rights to manage and so on).

Idea: adopt the same behaviour  as in the `uptime` service,
and add  a timeout in `--warning`  (eg `2d`).

So, after a parameter change, the service would reset itself after some duration. From Icinga, it would be only necessary to ack the service.

BTW, the service never  raises a Critical although the doc says so, and the code seems to hint that it was supposed to raise a Critical when a restart is pending. This last status should not be cleared; it will become a simple Warning after the restart, and clears itself if the timeout is set.

blogh:
With this patch Warning is raised when the time since configuration change
is less than the warning threshold. Obviously these alerts will disappear from
themselves once enough time has passed. This doesn't apply for configuration 
changes with pending restart warnings.

Eg:

```
./check_pgactivity -s settings -h /var/run/postgresql -p 5433  -w 1s
```

Closes #405

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -7771,17 +7771,15 @@ PARAM_LOOP: foreach my $row (@rs) {
     $expected_result = 2 if scalar @long_msg;   # Regular warning
     $expected_result = 1 if $pending_count > 0; # Pending restart
     $this_msg = join('\0', @long_msg);
-    if ( not defined $settings_laste{'date'} || not defined $settings_laste{'msg'}  ) {
-        $settings_laste{'msg'} = "";
-        $settings_laste{'date'} = $now;
-    }
+    $settings_lastemsg = $settings_laste{'msg'} // "";
+    $settings_lastedate = $settings_laste{'date'} // $now;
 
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -7670,9 +7670,12 @@ again with the argument C<--save>.
 
 No perfdata.
 
-Critical and Warning thresholds are ignored.
+The Critical thresholds is ignored.
 
-A Critical is raised if at least one parameter changed.
+Warning is raised when the time since configuration change is less than the
```
Krysztophe: Nitpicking: since configuration change or since the 1st time a change was detected?

blogh: Since the first time this particular change was detected.

blogh:
rebase

Issue #411, opened 2026-02-19
pgss_dealloc: fails with `--dbservice`  /  service target must have pg_stat_statements


Krysztophe:
1) With `--dbservice`, it complains that there is no DB:
```
FATAL: you must give one database with service "pgss_dealloc".
```
Patch is coming to allow that.

2) If the database in the service is `template1` or `postgres` as often,
it is not possible to query a `pg_stat_statements` installed in another DB.

That may be a limit. I don't know of a work-around.

Perhaps is it possible to `--dbinclude` or another parameter to make another connection once connected, but this may be overkill.

gleu:
I've merged the first part. If you still think the second part remains interesting, open a new issue, and work on a PR. Thank you.

PR #412, opened 2026-02-19
pgss_dealloc: allow --dbservice + doc

Krysztophe:
Cf #411

* modify a check to allow `pgss_dealloc` to use `--dbservice`
* modify the error message
* add doc for the reprequisites

Does not solve the problem of connection with a service when pgss is installed in another db.


gleu:
I've merged the code, and don't know why github didn't get it (probably the rebase I did).

Issue #389, opened 2025-10-15
Remaining unit tests to implement


frost242:
Here is the complete list of services. PR #387 implements a few more tests. The aim of this list is to provide an overview of the remaining unit tests to achieve :

- [x] 	archive_folder
- [x] 	archiver
- [x] 	autovacuum
- [x] 	backends
- [x] 	backends_status
- [x] 	backup_label_age
- [x] 	bgwriter
- [x] 	btree_bloat
- [x] 	checksum_errors
- [x] 	commit_ratio
- [x] 	configuration
- [x] 	connection
- [x] 	custom_query
- [x] 	database_size
- [x] 	extensions_versions
- [x] 	hit_ratio
- [x] 	hot_standby_delta
- [x] 	hugepages
- [x] 	invalid_indexes
- [x] 	is_hot_standby
- [x] 	is_master
- [x] 	is_replay_paused
- [x] 	last_analyze
- [x] 	last_vacuum
- [x] 	locks
- [x] 	longest_query
- [x] 	max_freeze_age
- [x] 	minor_version
- [x] 	oldest_2pc
- [x] 	oldest_idlexact
- [x] 	oldest_xmin
- [ ] 	pg_dump_backup
- [x] 	pga_version
- [x] 	pgdata_permission
- [x] 	replication_slots
- [x] 	sequences_exhausted
- [x] 	session_stats
- [x] 	settings
- [x] 	stat_snapshot_age
- [x] 	streaming_delta
- [x] 	table_bloat
- [x] 	table_unlogged
- [x] 	temp_files
- [x] 	uptime
- [x] 	wal_files

frost242:
So, PR #387 should have completed a minimal coverage for check_pgactivity. pg_dump_backup is still on hold, but does not depend on any PostgreSQL functionnality.

anayrat:
Thanks for this work!
There is something weird with temp_files test.
In the Github actions tests,  we dump the logs in order to facilitate troubleshooting. 
But, the temp_files tests seems to generate hundred of megabyte of logs : 
```
2025-10-18T15:43:20.1877802Z # Running: ./check_pgactivity --service temp_files --username runner --format human --dbname template1
2025-10-18T15:43:20.1878100Z ok 4 - basic check status (got 0 vs expected 0)
2025-10-18T15:43:20.1878270Z ok 5 - basic check stdout /(?^m:^Service  *: POSTGRES_TEMP_FILES$)/
[...]

Issue #367, opened 2023-11-30
Last vacuum of TOAST tables  & Last vacuum/analyze of pg_catalog tables


Krysztophe:
[As found by  @gleu ](https://github.com/OPMDG/check_pgactivity/issues/282#issuecomment-766898307), the current implementation of `last_maintenance` uses `pg_stat_user_tables` which gets rid of `pg_catalog` and TOAST tables: 

 ```sql
\sv+ pg_stat_user_tables 
CREATE OR REPLACE VIEW pg_catalog.pg_stat_user_tables AS
SELECT pg_stat_all_tables.relid,
…
FROM pg_stat_all_tables
WHERE (pg_stat_all_tables.schemaname
<> ALL (ARRAY['pg_catalog'::name, 'information_schema'::name]))
AND pg_stat_all_tables.schemaname !~ '^pg_toast'::text
``` 

So :

  * This is not an issue for `last_analyze` on TOAST tables, as they are never analyzed by themselves
  * We may miss a failing or missing VACUUM on TOAST tables (especially now that VACUUM can explicitly exclude them).
  *  The VACUUM and ANALYZE on `pg_catalog` tables are not monitored, although they should, as they are a bit critical. 

Suggestion : base the queries on `pg_stat_all_tables`, exclude only `information_schema`. For `last_analyze`, exclude `pg_toast_%` too.

Could be made at the same time than #365

gleu:
PR is merged, issue is closed :)

PR #417, opened 2026-03-29
Checks pg_catalog tables for last_*

gleu:


gleu:
@anayrat If you have some time, can you check this? I have no idea why it fails on v12, v13, and v14 in the CI, but works on my laptop.

anayrat:
What is the purpose of this PR ? Just to understand.

gleu:
This fixes the issue described in #367. In other words, it also checks catalog tables. As simple as a s/pg_stat_user_tables/pg_stat_all_tables/.

It definitely needs a rebase with the last merge, but that's not the issue right now. My main problem here is: why does it work with "prove t" on my laptop, and not on the CI? (I've only checked v14 as that's the only RPM available for the releases with a failed status.)

anayrat:
I don't know why. I am not able to run prove on my VM. Tests fails randomly on my VM https://github.com/OPMDG/check_pgactivity/issues/389#issuecomment-3421712112

It seems we have tables from system catalog not vacuumed despite the `VACUUM ANALYZE` just before.

anayrat:
Ok, it seems stats are not fully consistent. Maybe stats collector is really slow on CI runner. I added a pause for one second after the `VACUUM ANALYZE` and it is all green now \o/

gleu:
That seems quite fragile. Anyway, if it's working, I won't complain :)

Thank you.

anayrat:
Yes, I don't like adding sleep in tests. But it shows stats tables are not fully consistent, it is good to know :)

gleu:
Merged, thank you.

PR #420, opened 2026-05-04
Add a unused_indexes service

### LINKED ISSUE #32: Check for unused indexes
Is it possible to add a check for unused indexes?

Josh Berkus had a nice blog post about that in:
http://www.databasesoup.com/2014/05/new-finding-unused-indexes-query.html

Probably also a check for sequencial scans on a table since last execution of check.

gleu:
Basic fix for #32

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -247,6 +247,10 @@ my %services = (
         'sub'  => \&check_invalid_indexes,
         'desc' => 'Check for invalid indexes.'
     },
+        'unused_indexes' => {
+        'sub'  => \&check_unused_indexes,
+        'desc' => 'Check for unused indexes.'
+    },
     'is_master' => {
         'sub'  => \&check_is_master,
         'desc' => 'Check if cluster is in production.'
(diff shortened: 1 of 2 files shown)


frost242:
Hello,

Thanks for taking some time to fix and add new things on check_pgactivity. Regarding this service, I don't agree that it's a feature that we want. Checking for unused indexes is challenging and can lead to "false positives" : think about an index mandatory in a monthly or yearly batch. Think on an index backing an unique constrainte (PK or not).

A simple test leads to a false positive :
```
[postgres]# create table t1 (i integer primary key);
CREATE TABLE
[postgres]# insert into t1 values (1);
INSERT 0 1
[postgres]# insert into t1 values (2);
INSERT 0 1
[postgres]# insert into t1 values (3);
INSERT 0 1
```

The query embedded in the script will make this service bail out about the index supporting the PK.

Krysztophe:
I agree that all indexes for PKs and UNIQUE keys should be excluded.

As for indexes on FKs, this is more difficult. They should be present by default there until proven useless, this service could help to detect them.

You need a big warning that an index may be useful once a year, but this service could be useful if not used blindly.

gleu:
I added support for checking primary key and unique constraints indexes.

It may bring false positive, but many other services do too. I still think it's worth it.

PR #426, opened 2026-06-03
check_last_maintenance : filter out small tables

yhuelf:
What is the purpose of the `last_vacuum` and `last_analyze` probes?

* To be alerted when statistics are reset following a crash, a failover, etc.
* To detect large tables that are no longer being processed frequently enough due to the default autovacuum settings, which are not always well-suited for large tables.

Alerts for small tables therefore constitute unnecessary noise.

### DIFF
--- a/README
+++ b/README
@@ -772,7 +772,7 @@ COMPATIBILITY
 
         You can use multiple "--exclude REGEX" parameters.
 
-        This service supports a "--analyze_table_min_size" parameter to
+        This service supports a "--analyze-table-min-size" parameter to
         exclude small relations. We use the `pg_relation_size()` function
         to filter out those relations that are smaller than this value,
         because the tables and their associated toast tables are processed
@@ -816,7 +816,7 @@ COMPATIBILITY
 
         You can use multiple "--exclude REGEX" parameters.
 
-        This service supports a "--vacuum_table_min_size" parameter to
+        This service supports a "--vacuum-table-min-size" parameter to
         exclude small relations. We use the `pg_relation_size()` function
         to filter out those relations that are smaller than this value,
         because the tables and their associated toast tables are processed
--- a/README
+++ b/README
@@ -773,9 +773,10 @@ COMPATIBILITY
         You can use multiple "--exclude REGEX" parameters.
 
         This service supports a "--analyze_table_min_size" parameter to
-        exclude small relations. The value must be a positive integer
-        representing the threshold in kilobytes below which the relation
-        is excluded. This defaults to 1024 (1MB).
+        exclude small relations. We use the `pg_relation_size()` function
+        to filter out those relations that are smaller than this value,
+        because the tables and their associated toast tables are processed
+        separately by autovacuum, and by this probe. This defaults to 1MB.
 
         Required privileges: unprivileged role able to log in all databases.
 
@@ -816,9 +817,10 @@ COMPATIBILITY
         You can use multiple "--exclude REGEX" parameters.
 
         This service supports a "--vacuum_table_min_size" parameter to
-        exclude small relations. The value must be a positive integer
-        representing the threshold in kilobytes below which the relation
-        is excluded. This defaults to 10240 (10MB).
+        exclude small relations. We use the `pg_relation_size()` function
+        to filter out those relations that are smaller than this value,
+        because the tables and their associated toast tables are processed
+        separately by autovacuum, and by this probe. This defaults to 2MB.
 
         Required privileges: unprivileged role able to log in all databases.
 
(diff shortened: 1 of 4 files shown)


README
```diff
@@ -772,6 +772,11 @@ COMPATIBILITY
 
         You can use multiple "--exclude REGEX" parameters.
 
+        This service supports a "--analyze_table_min_size" parameter to
+        exclude small relations. The value must be a positive integer
+        representing the threshold in kilobytes below which the relation
+        is excluded. This defaults to 1024 (1MB).
```
blogh: We should specify that we are talking about the "main fork" and maybe reflect
this in the parameter name.

README
```diff
@@ -810,6 +815,11 @@ COMPATIBILITY
 
         You can use multiple "--exclude REGEX" parameters.
 
+        This service supports a "--vacuum_table_min_size" parameter to
+        exclude small relations. The value must be a positive integer
+        representing the threshold in kilobytes below which the relation
+        is excluded. This defaults to 10240 (10MB).
```
blogh: same as `--analyze_table_min_size`

README
blogh: I would expect [units](https://github.com/OPMDG/check_pgactivity/blob/8b7917127a56b1592b208e8b3b979946e461bc33/check_pgactivity#L894) to work here. 

I don't know if that's necessary, we will likely filter small table so the value should be easy to 
read whthout it. But it would be consistent with other sizes.

blogh (commented):
It's a great addition.

README
Krysztophe: main fork only? toast not included?

check_pgactivity
```diff
@@ -503,6 +503,8 @@ my %args = (
     'wal_buffers'           => undef,
     'checkpoint_segments'   => undef,
     'effective_cache_size'  => undef,
+    'vacuum_table_min_size' => 10 * 1024,
```
Krysztophe: 10 MB is already too much I think.

Most small tables that gives false alarms are only a few blocks wide.

We can discuss at length the right limit, but since  we had service that was much too sensitive, perhaps shall we raise the bar not so high.

README
yhuelf: > main fork only? toast not included?

Yes because the toast table is processed separately by autovacuum. The main fork could be only 8kB, 100 tuples, never vacuumed, while the toast table can be 100MB, 50000 tuples, already vacuumed. If you use `pg_table_size` instead of `pg_relation_size`, you will get an alert for the main fork, and we don't want that.

check_pgactivity
yhuelf: what about 2 MB ?

yhuelf: I think 1000 tuples would be a better threshold (`autovacuum_vacuum_insert_threshold` default value).

A 2MB table should contains at least that number of tuples.

README
yhuelf: > I would expect [units](https://github.com/OPMDG/check_pgactivity/blob/8b7917127a56b1592b208e8b3b979946e461bc33/check_pgactivity#L894) to work here.

fixed in 6bad3f316db1

t/01-last_analyze.t
```diff
@@ -50,6 +50,7 @@ sleep(1);
 
 $node->command_checks_all( [
     './check_pgactivity', '--service'  => 'last_analyze',
+                          '--analyze_table_min_size' => 0,
```
gleu: You should only use dash in a CLI option, so no underline (meaning --analyze-table-min-size).

t/01-last_analyze.t
```diff
@@ -106,6 +107,7 @@ SKIP: {
...
```
gleu: Same here.

t/01-last_analyze.t
```diff
@@ -145,6 +147,7 @@ push @stdout, (
...
```
gleu: Actually, same everywhere :)

t/01-last_vacuum.t
```diff
@@ -50,6 +50,7 @@ sleep(1);
 
 $node->command_checks_all( [
     './check_pgactivity', '--service'  => 'last_vacuum',
+                          '--vacuum_table_min_size' => 0,
```
gleu: Ditto (--vacuum-table-min-size).

gleu (changes requested):
I'm OK with the idea, but the CLI option' names need some changes (no underscore, but dashes).

gleu (approved):
Fine with me.

check_pgactivity
Krysztophe: ok for 2 Mo

gleu:
Merged (see 8a7e47d). Thank you.

PR #427, opened 2026-06-08
Fix relation_size documentation

ioguix:
Following our discussion offline about this documentation.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -8653,28 +8653,25 @@ sub check_table_unlogged {
 
 Check relations' size.
 
-Without C<--critical>  or C<--warning> parameters, this service attempts to
-fetch all relations' size.
+Critical and Warning thresholds are optional. An alert is raised if a
+relation's size is above the warning or critical level. They only accept a
+raw number or a size.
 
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -8653,28 +8653,25 @@ sub check_table_unlogged {
...
-You can use multiple C<--exclude REGEX>  parameters.
-
-Perfdata will return the number of relations per database.
+You can use multiple C<--exclude REGEX> parameters.
 
-A list of the relations and their sizes will be returned after the
-perfdata. This list contains the fully qualified relation name. If
-C<--exclude REGEX> is set, the number of excluded relations is returned.
+Perfdata returns the size of all relations per database. This list use the
+fully qualified of each relation name. If C<--exclude REGEX> is set, the number
```
gleu: It should be "fully qualified name of each relation"

gleu (changes requested):
Sounds good to me except one sentence.

check_pgactivity
ioguix: done in new commit

gleu (approved):
Fine with me.

PR #432, opened 2026-08-25
Replace gettimeofday() with time()

gleu:


frost242:
Hello,

I think we should merge this PR. But I wonder why this commit also modifies the rpm spec file to add "Requires: perl-FindBin". Sorry, I miss some context and it's been years since I last used a RHEL-based distro.

gleu:
Yeah, you're right. It probably shouldn't. We need this new requirement. I couldn't add the RPM with this one.

gleu:
Added a commit to remove the requirement in the spec file.

frost242:
Thanks. so let's merge it. Thanks @gleu !

gleu:
You're welcome, and thank you.

Issue #429, opened 2026-06-11
New GUCs to check in configuration service


gleu:
@Krysztophe was interested with data_checksums. I think data_sync_retry would be good too (https://thebuild.com/blog/all-your-gucs-in-a-row-datasyncretry/). Do we want more checks?

Krysztophe:
If you ask me, below of the summary of what I'd like to check.

Note that we cannot add as many `--no-check*-` flags, something else should be used.

* Currently we check:
  - `fsync = on`
  - `autovacuum  = on`
  - some minimal values of `work_mem`, `shared_buffers`, `checkpoint_segments`, `wal_buffers`, `effective_cache_size`
  -  `enabled_*` (except 2)
  - `stats_row_level = on`   (_remember this one? me neither_) or `track_counts=on` (8.3)

* Bug to correct:
  - `max_wal_size` was forgotten when `checkpoint_segments` disappeared in 9.5  ( `min_wal_size` may be forgotten)

* Most important things that I'd like to add:
  - `data_checksums = on`
  - `data_sync_retry = off`
  - `allow_system_table_mods =off`
  -  `full_page_writes = on`  (except ZFS ?)
  - `zero_damaged_pages = off`

* I think it would be worth to check less important parameters whose values are now recommended.
Error messages may be a challenge to write.
  -  `wal_compression != off`
  - `lc_messages = C` or `en%`
  -  `password_encryption = scram-sha-256`
  - `remove_temp_files_after_crash = on`   (I'm discovering its existence)
  - `checkpoint_timeout` >= 5min
  - `checkpoint_completion_target` >= 0.8 eg
  - `server_encoding = UTF8`
  - `deadlock_timeout` >=0.5 s  (?)
  - `autovacuum_worker_slots` >= 3  (default : 16)
  - `default_statistics_target` <=1000
  -  `default_toast_compression != pglz`
  - `effective_io_concurrency  !=1`  (the default is now 16, and IIRC 1 is the worst possible value)
  - `autovacuum_*_threshold > 100 000` when `autovacuum_*_scale_factor=0` ?  (to avoid an hysterical autovacuum, as I've seen once)
  - `from/join_collapse_limit` <= 12 (too dangerous cluster-wide to go higher for OLTP)
  - `ssl = on` 
  - some `log_*` parameters?

We could go as far as adding `--ssd`, `--oltp`, `--olap` parameters.

gleu:
debug_assertions should also be disabled on a production node.

[...]

Issue #283, opened 2021-02-19
Detect PG configuration difference between 2 servers


Krysztophe:
Original use case: configuration was changed on the primary, not copied on the replica. After the restart, the replica does not even start (different max_connections). Or the replica behaves differently than the old primary after a failover.

''settings'' or ''configuration'' services raise alerts when the conf changes or there is something pending, but this is not enough.

Need: a check that raises an alert when the configuration is or will be different on 2 (or more!) servers. These are not always primary/replicas (eg. sharding). Critical could be raised for incoherences preventing a restart (eg. max_connections).

Exclusion/exclusion of parameters would allow some flexibility. Some parameters could be excluded by default (eg. cluster_name, perhaps log_* ?)

Option 1) Compare ''pg_settings'' values. This works only after a ''pg_reload_conf()'', so some configuration can be forgotten . This does not work on a replica which is already down.

Option 2) Get the parameters from postgresql.conf/auto.conf.  This is painful due to included files. ''postgres -C'' is very nice for this but need to be provided with the parameter name; AFAIK you cannot get the whole configuration this way. And this needs a filesystem access, some rights, and to define binary & PGDATA paths.

Perhaps is there place for 2 services?  compare_active_conf and compare_restart_conf ?

Krysztophe:
For option 2, `pg_file_settings`  would be helpful

gleu:
Merged.

PR #433, opened 2026-08-25
Check for more GUCs configuration

gleu:
This is a PR for issue #429. It handles right now: fix for max_wal_size, adds the most important settings Christophe talked about, and debug_assertions.

Some GUCs are missing but it needs more discussion in the issue.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -503,6 +503,7 @@ my %args = (
     'shared_buffers'        => undef,
     'wal_buffers'           => undef,
     'checkpoint_segments'   => undef,
+    'max_wal_size'          => undef,
     'effective_cache_size'  => undef,
     'vacuum-table-min-size' => '2mb',
     'analyze-table-min-size'=> '1mb',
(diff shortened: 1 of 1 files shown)


check_pgactivity
```diff
@@ -3872,8 +3873,9 @@ Warning and Critical thresholds are ignored.
 
 Specific parameters are :
 C<--work_mem>, C<--maintenance_work_mem>, C<--shared_buffers>,C<--wal_buffers>,
-C<--checkpoint_segments>, C<--effective_cache_size>, C<--no_check_autovacuum>,
-C<--no_check_fsync>, C<--no_check_enable>, C<--no_check_track_counts>.
+C<--checkpoint_segments>, C<--max_wal_size>, C<--effective_cache_size>,
+C<--no_check_autovacuum>, C<--no_check_fsync>, C<--no_check_enable>,
+C<--no_check_track_counts>.
```
Krysztophe: Easy to understand but painful to maintain?

Perhaps something like `--no_check='maintenance_work_mem,checkpoint_segments'` would be better?

gleu: Yeah, probably. I'm not fluent enough in Perl to write this, so I guess the best thing to do is to open another issue with this feature request, and finish this one with what it already has (and pray that someone with more Perl knowledge grabs the new issue).

Krysztophe: I'd like to have the time to try...

The problem is that all this parameters cannot be removed once they're there (or you break compatibility).

I'm wondering if so many different parameters may be a problem on Icinga/Nagios side.

gleu: I only added --max_wal_size. So the compatibility issue is already there. One way to avoid it is to let the ones already there, mark them obsolete, and add the new --no_check option. Or better --no-check.

Krysztophe: I've just thought that the problem will be the same for https://github.com/OPMDG/check_pgactivity/issues/283, and --no-check will be useful there too.

gleu: Still, better on a new issue, specific to this new CLI option.

Krysztophe: BTW, we do not need a --no-check, the --exclude regexes are ok (see many examples in the code)

gleu: You're right, and it was quite easy to do. I added a commit to handle this.

gleu:
Now that we have --exclude, the newly added --no-max_wal_size should be expurged from the patch. I'll work on this today.

gleu:
> Now that we have --exclude, the newly added --no-max_wal_size should be expurged from the patch. I'll work on this today.

Scratch that, there's no --no-max_wal_size. So no work to do there.

check_pgactivity
Krysztophe: tested, works for me

Krysztophe:
I've played with it on my test and personal instances : that works.

Only problem: it raises a CRITICAL while many problems are not immediately dangerous (wal_compression, lc_messages…) and deserves a WARNING.

Perhaps add a 3rd column in the query that would return 1 or 2, and a UNION testing different criteria ?

I could do this afternoon.

gleu:
You can try. I don't see why you need a third column, but your call :)

Krysztophe:
I've pushed  627a2d92c498b00546c66d6a5c10ac27c99d1f50  to raise a Critical or Warning according to the parameter. We can always discuss the case of each GUC.

I've tested from 8.2 to 19, it seems to work.

Update of the doc:  627a2d92c498b00546c66d6a5c10ac27c99d1f50

gleu:
Sounds good to me.

gleu:
@Krysztophe do you still have some work left on this or can I merge this?

Krysztophe:
For me it's done and awaiting returns. Merge if you think it's ready enough.

gleu:
Done.

gleu:
And thank for all your work on this issue, @Krysztophe !

PR #434, opened 2026-08-25
archive_mode isn't a boolean since 9.5

gleu:


frost242:
lgtm ! thanks a lot for this fix !

frost242:
Next step is probably to handle archive_mode = 'always'.

PR #436, opened 2026-08-26
Add option --ok-on-standby

### LINKED ISSUE #359: Last Vacuum Always Errors on Hot Standby Cluster

gleu:
For last_vacuum and last_analyze services, returns OK on a standby if this new option is used. Otherwise, returns UNKNOWN.

Fixes #359.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -525,7 +525,8 @@ my %args = (
     'uid'                   => undef,
     'with-hugepages'        => undef,
     'schexclude'            => undef,
-    'relexclude'            => undef
+    'relexclude'            => undef,
+    'ok-on-standby'         => 0
 );
 
 # Set name of the program without path*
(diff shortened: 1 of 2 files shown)


frost242:
looks good. Do you have some time to add another unit test for this specific case ? Let me know if not, I'll take a look when possible.

gleu:
I'll work on this this afternoon.

gleu:
Done, just added two unit tests for the new CLI option. Sounds good?

frost242:
Very good, thanks a lot. let's merge this PR.

gleu:
Thanks. You're quite active these days, I love this!

PR #439, opened 2026-08-29
Configuration difference

Krysztophe:
See issue #283 (option 1).

This is the 1st draft of a new service `configuration_difference` checks that two servers have the same configuration.
```
 ./check_pgactivity --dbservice=defo14,defo14-2  -s configuration_difference --ignore-paths 
POSTGRES_CONFIGURATION_DIFFERENCE WARNING: [defo14-2] : archive_command=(disabled) (≠ /bin/true), cluster_name=14/defo2 (≠ 14/defo), log_connections=off (≠ on), log_disconnections=off (≠ on), port=14012 (≠ 14001), shared_preload_libraries= (≠ pg_stat_statements), wal_keep_size=0 (≠ 100)
```

By default, it assumes that the 1st server in `--dbservice server1,server2` is a primary
and the others are secondaries that must be fully identical.

The check is based on `pg_settings`, as files in `pg_file_settings` are not always applied.

The monitoring user must not have some configuration that hides the server parameters, but usually a monitoring user has nothing important.

By default it ignores a few parameters that will almost always differ on a secondary:
`archive_mode,primary_conninfo,primary_slot_name`
(`in_hot_standby` and `transaction_read_only` too).
I've left `archive_command` and `restore_command`, they should be the same on a secondary ready for a failover.

`--identical`  allows to override this (sharding context?)

Some parameters will be always ignored (I've thought only of `effective_wal_level` in v19)

I'm probably not the only one with many test instances on the same server,
so I've added `--ignore-paths` to ignore all differences in paths.
I'm not sure about ignoring `cluster_name` too.

`--exclude`  works:
```
 --exclude '(.*parallel.*|cluster_name|auto_explain.*|pg_stat_statements.*|work_mem)'
```

Compatibility: no minimum version as `name`,`setting`  are there from 7.4 at least (I didn't test that far). If `pending_restart` may be of interest (add it in the message?), that would limit to 9.5+.

No perfdata : I don't see the point.

I'm not too sure about the output message format.

Returns a :yellow_circle: Warning if some parameters are different:
I've not thought of a case where a  small parameter difference is always a major problem  (or the secondary is already down).

Known limits:
- roles configurations are ignored (at least for now; not sure if this service is the place) ;
- the connection role may hide the "real" configuration as said above;
- custom parameters (eg. `demo.toto`) cannot be checked if not declared by their extension (they do not appear in pg_settings).

There is no test because I have no clue how it works and it's already late.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -535,6 +535,7 @@ my %args = (
     'schexclude'            => undef,
     'relexclude'            => undef,
     'ok-on-standby'         => 0,
+    'identical'             => 0,
 );
 
 # Set name of the program without path*
@@ -3991,15 +3992,16 @@ LOOP: foreach my $setting (@rc) {
 =item B<configuration_difference>
 
 Check that the configuration is the same between two servers (according to pg_settings)
-By default, checks that the 1st server is a primary
-and the next one secondary that must be fully identical.
+By default, assumes that the 1st server is a primary
+and the others are secondaries that must be fully identical.
 
-Give the connection parameters with 2 or more services,
+Give the connections parameters with 2 or more services,
 eg. with --dbservices=pg1,pg2
 The first one is taken as the reference (usually a primary).
 
-By default it ignores archive_mode,primary_conninfo,primary_slot_name.
-Force the check with C<--identical>.
+By default it ignores a few parameters that will differ on a secondary:
+archive_mode,primary_conninfo,primary_slot_name.
+Force the check of all parameters with C<--identical>.
 
 This service supports the argument C<--exclude REGEX> to exclude parameters
 matching the expression. You can use multiple C<--exclude REGEX> arguments.
@@ -4007,21 +4009,19 @@ Eg : --exclude '(.*parallel.*|cluster_name|auto_explain.*|pg_stat_statements.*|w
 
 No perfdata
 
-Returns a Warning if some parameters are different.
+Returns a Warning if some parameters are different
 
 Required privileges: unprivileged role
 
-Known issues: roles configuration is ignored ;
-the connection role may hide the real configuration ;
-custom parameters (eg. demo.toto) are ignored if not declared by their extension.
+Known issues: roles configurations are ignored ;
+the connection role may hide the "real" configuration ;
+custom parameters (eg. demo.toto) cannot be checkd if not declared by their extension.
 
 =cut
 
 sub check_configuration_difference {
     my @perfdata;
     my @msg;
-    my @msg_crit;
-    my @msg_warn;
     my @hosts;
     my @cfg;
     my %refserver;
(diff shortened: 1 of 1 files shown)


Krysztophe:
To deal with tcp_keepalives* & co that are  not available through a socket, I changed the query and the setting changes to N/A if this case, and this is ignored.

check_pgactivity
```diff
@@ -2509,7 +2516,7 @@ C<undefined>, C<disabled> and C<insufficient privilege>.
 B<insufficient privilege> appears when you are not allowed to see the statuses
 of other connections.
 
-This service supports the argument C<--exclude REGEX> to exclude queries
+This service supports the argument C<--exc REGEX> to exclude queries
```
gleu: I guess this is an unwanted change?

check_pgactivity
```diff
@@ -3942,6 +3949,136 @@ DB_LOOP:    foreach my $setting (@rc) {
...
+Returns a Warning if some parameters are different
+
+Required privileges: unprivileged role
+
+Known issues:
+TCP configuration can be checked only when two instances
+are connected by TCP (no socket);
+roles configurations are ignored;
+the connection role may hide the "real" configuration;
+custom parameters (eg. demo.toto) cannot be checkd if not declared by their extension.
```
gleu:
```suggestion
custom parameters (eg. demo.foo) cannot be checked if not declared by their extension.
```

check_pgactivity
```diff
@@ -3942,6 +3949,136 @@ DB_LOOP:    foreach my $setting (@rc) {
...
+    my $mess            = "";
+    my $setting_ref     = "";
+    my @exclude_regexes = @{ $args{'exclude'} };
+    my $identical       = $args{'identical'} ;
+    my $ignore_paths    = $args{'ignore-paths'} ;
+    # will probably or always different on a primary & its secondary
+    my @ignored_if_secondary = ( 'archive_mode', 'in_hot_standby',
+                                 'primary_conninfo', 'primary_slot_name',
+                                 'transaction_read_only' ) ;
+    my @ignored_paths = ( 'config_file','data_directory','external_pid_file',
```
gleu, line 4050: Why do you call it _path? they aren't paths. I'm probably missing something.

gleu (changes requested):
This was a quick review, without tests. 3 comments right now.

check_pgactivity
Krysztophe: Ooops! I'll correct

check_pgactivity
Krysztophe: These parameters all contain paths, and change if the primary is on `/var/lib/postgresql/18/prod1` and the secondary on `/var/lib/postgresql/18/prod2`, for example.

An idea for a better name? `ignore_because_of_path`?

check_pgactivity
Krysztophe: done

check_pgactivity
gleu: Oh OK, understood. So, current name is fine with me.

gleu:
Sounds good for a merge. Be careful that I pushed a commit that reorder the service list. I won't be surprised if you get a conflict.

check_pgactivity
Krysztophe: But it needs a comment

```suggestion
    # the parameters contain paths that may differ on two different servers
    # used by --ignore-paths
    my @ignored_paths = ( 'config_file','data_directory','external_pid_file',
```

gleu: Sounds good!

Krysztophe:
Rebasing...

Krysztophe:
Rebased

gleu:
Thanks @Krysztophe !

Issue #157, opened 2018-01-26
Report settings in error


frost242:
Hello,

Since release 9.5, PostgreSQL offers the pg_file_settings view to detect errors in the current configuration files.
Also, PostgreSQL 10 now offers the pg_hba_file_rules view to detect errors in the pg_hba.conf file.

Errors in both files can prevent PostgreSQL to restart. We should warn about such thing. Maybe a check_settings.

We already warn if we encounter a setting that is pending a restart.

gleu:
Sounds like a new service would be better for this one. file_settings would be a nice name.

gleu:
MR #450 is my attempt to add this service. Still draft, because I want to add perfdata per configuration file.

gleu:
It also needs a unit test.

gleu:
Perfdata are now available per database:

```
➜ ./check_pgactivity -s file_settings -F human
Service        : POSTGRES_FILE_SETTINGS
Returns        : 2 (CRITICAL)
Message        : 4 errors
Long message   : file /var/lib/pgsql/19/data/postgresql.conf, line 934, setting array_nulls, error: setting could not be applied
Long message   : file /var/lib/pgsql/19/data/postgresql.conf, line 935, setting exit_on_error, error: setting could not be applied
Long message   : file /var/lib/pgsql/pg_hba.conf, line 6, error: invalid authentication method "::1/128"
Long message   : file /var/lib/pgsql/pg_ident.conf, line 5, error: missing entry at end of line
Perfdata       : total errors=4
Perfdata       : errors in /var/lib/pgsql/19/data/postgresql.conf=2
Perfdata       : errors in /var/lib/pgsql/pg_ident.conf=1
Perfdata       : errors in /var/lib/pgsql/pg_hba.conf=1
```

gleu:
So, I was too quick, and didn't think it through. We already have issue #374 and PR #440 for errors in pg_hba.conf. After a quick conversation with some colleagues, we agreed on having a service per configuration file: settings for postgresql.conf files, hba_settings for pg_hba.conf files, ident_settings for pg_ident.conf files. I'm going to rename the service in PR #440, and use this PR to checks for error on the settings service. I'll build another PR for the new ident_settings service.

Let's go.

gleu:
Done for PR #440.

gleu:
PR #450 has the changes for the settings service.

gleu:
PR #451 adds the new ident_settings service.

[...]

PR #440, opened 2026-08-31
New pg_hba_issues service

### LINKED ISSUE #374: Service checking most obvious security mistakes in pg_hba.conf
We could check some obvious security problems in  pg_hba_file_rules , eg:

* existence of trust or password method (CRITICAL)
* md5 instead of scram (WARNING ?)
* syntax error in pg_hba.conf (CRITICAL)

Some others could be activated with options (host instead of hostnossl, 0.0.0.0 address…), and a perfectly configurable setup is probably an illusion, but a very basic service would already be already useful.

gleu:
It reports the number of errors reported in pg_hba_file_rules, and the number of rules with the trust, password, and md5 authentication methods. These methods are not recommended.

Warning and threshold values could be supplied, but this isn't recommended. All issues should be reported.

Fixes #374.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -180,6 +180,10 @@ my %services = (
         'sub'  => \&check_extensions_versions,
         'desc' => 'Check that installed extensions are up-to-date.'
     },
+    'hba_settings' => {
+        'sub'  => \&check_hba_settings,
+        'desc' => 'Check issues in HBA configuration.'
+    },
     'hit_ratio' => {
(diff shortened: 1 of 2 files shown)


gleu:
It needs a unit test. I'll work on this today.

gleu:
Just added a simple unit test. Might need more work but the basics are here.

gleu:
Service has been renamed hba_settings. See #157 for details.

PR #441, opened 2026-08-31
backup_label_age : Critical on PG15+ if backup_label exists

### LINKED ISSUE #322: backup_label : return immediately CRITICAL on secondary, or when pg_is_in_backup() = false
- A backup_label on a secondary server is not normal and dangerous.
The service should return CRITICAL immediately.

- If pg_is_backup () is false, it should return CRITICAL too
(that would cover the previous case).

Krysztophe:
Backup_label is not supposed to exist on a running PG15+ instance,
  as the concurrent backups do not exist anymore. Its presence still
  prevents a restart, so raise à Critical immediately.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -2907,14 +2907,23 @@ sub check_checksum_errors {
 
 =item B<backup_label_age> (8.1+)
 
-Check the age of the backup label file.
+Check the presence of backup label file (≥ PG15),
+or raise an alarm if it's too old (≤ PG14).
+
+Until PostgreSQL 14, a backup_label appears while a concurrent physical
+backup is running.
+Critical or Warning are raised if the age is above the thresholds
+(only intervals, eg. 1h30m25s).
+
+Thresholds are ignored on PostgreSQL 15 and above :
+the presence of a backup_label is an immediate failure (Critical),
+as this file should never be present on a running system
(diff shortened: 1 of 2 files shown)


Krysztophe:
Test updated (perhaps not in the cleanest way)

Krysztophe:
Tests updated and green ;  rebased ; ready for review.

check_pgactivity
```diff
@@ -2899,14 +2899,23 @@ sub check_checksum_errors {
...
+or raise an alarm if it's too old (≤ PG14).
 
-Perfdata returns the age of the backup_label file, -1 if not present.
+Until PostgreSQL 14, a backup_label appears while a concurrent physical
+backup is running.
+Critical or Warning are raised if the age is above the thresholds
+(only intervals, eg. 1h30m25s).
 
-Critical and Warning thresholds only accept an interval (eg. 1h30m25s).
+Threshold are ignored on PostgreSQL 15 and above :
```
gleu:
```suggestion
Thresholds are ignored on PostgreSQL 15 and above :
```

gleu (approved):
Sounds good to me, except the minor typo.

blogh (approved):
lgtm

gleu:
Merged, thanks.

PR #443, opened 2026-09-01
Add orphan_files service

gleu:
Right now, this is a simple POC. Still needs a better documentation, and a unit test.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -152,6 +152,10 @@ my %services = (
         'sub'  => \&check_checksum_errors,
         'desc' => 'Check data checksums errors.'
     },
+    'commit_ratio' => {
+        'sub'  => \&check_commit_ratio,
+        'desc' => 'Commit and rollback rate per second and commit ratio since last execution.'
+    },
     'configuration' => {
         'sub'  => \&check_configuration,
         'desc' => 'Check the most important settings.',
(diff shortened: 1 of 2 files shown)


check_pgactivity
```diff
@@ -6848,6 +6852,229 @@ sub check_oldest_xmin {
...
+
+This service supports both C<--dbexclude> and C<--dbinclude> parameters.
+
+Perfdata returns the number and total size of temp files found in
+the data directory and its tablespaces. They are aggregated by database.
+
+Critical and Warning thresholds are optional. They accept either a number
+of file (raw value), a size (unit is B<mandatory> to define a size) or both
+values separated by a comma. Thresholds are applied on current orphan files.
+
```
Krysztophe:
```suggestion
WARNING: Be very cautious when removing files. Always make a full backup before.
```

check_pgactivity
```diff
@@ -6848,6 +6852,229 @@ sub check_oldest_xmin {
...
+            $num_files++;
+            $size_files += $res->[2];
+        }
+
+        $total_num_files += $num_files;
+        $total_size_files += $size_files;
+
+        push @perfdata => [ "# files in $db", $num_files, 'File', @perf_flimits ];
+        push @perfdata => [ "Total size in $db", $size_files, 'B', @perf_limits ];
+    }
```
Krysztophe, line 7471: Something annoys me: you get the number and size of orphan files, but not their locations.

gleu: The first patch did that, so I can put it back. But I don't know in which part of the message. I won't use PerfData. I could use LongMsg but I have no idea of the consequences.

gleu: So, should I do that?

Krysztophe: I would say yes.
It is the 1st service that have very long messages, and this should be exceptional here.

Maybe you can use the `--detailed` flag (that already exists) to make everyone happy

gleu: Done with the latest commit. There's also a unit test.

gleu:
Rebase done.

PR #446, opened 2026-09-03
Parameters list: alphabetic reordering , add checks

Krysztophe:
- Alphabetic reordering of parameters declarations & checks in 3 places to make the addition of other parameters easier
- Added a missing check for --ok-on-standby  
(see the commits, the final result is unreadable in Github)


gleu:
Thank you, I've merged it.

PR #450, opened 2026-09-04
Adds handling of errors in settings service

gleu:
Check errors in file settings: postgresql.conf since 9.6, pg_hba.conf since 10, and pg_ident.conf since 15.

Perfdata contains the number of errors per files.

Thresholds, if any, are ignored.

Required privileges: unprivileged role.

### DIFF
--- a/check_pgactivity
+++ b/check_pgactivity
@@ -8135,11 +8135,11 @@ SLOTS_LOOP: foreach my $row (@rs) {
 =item B<settings> (9.0+)
 
 Check if the current settings have changed since they were stored in the
-service file.
+service file. Check also if there are errors in the configuration files.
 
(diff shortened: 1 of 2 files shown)


check_pgactivity
```diff
@@ -3622,6 +3626,122 @@ sub check_btree_bloat {
...
+            SELECT sourcefile, sourceline, name, error
+            FROM pg_file_settings
+            WHERE error IS NOT NULL
+            ORDER BY sourcefile, sourceline, name
+        },
+        $PG_VERSION_100 => q{
+            SELECT sourcefile, sourceline, name, error
+            FROM pg_file_settings
+            WHERE error IS NOT NULL
+            UNION
```
Krysztophe: UNION ALL !! (here and below)

